PR #4, opened 2021-06-30
Fix typos and clarify some terminology

mengelbart:


joerg-ott (approved):

joerg-ott:
Quick note: in a few places, we may need to speak of "RTP and RTCP packets" instead of just "RTP packets".
And I noticed: "A RTP packet" -> "An RTP packet"

joerg-ott (approved):

mengelbart:
> Quick note: in a few places, we may need to speak of "RTP and RTCP packets" instead of just "RTP packets".
> And I noticed: "A RTP packet" -> "An RTP packet"

Makes sense. I'll merge this PR first and then create another one for these changes.

PR #40, opened 2022-09-23
Add topology subsection

mengelbart:


aboba (commented):
"the translator either may need to rewrite"

[BA] Can we delete the "either"?  Also, you might want to clarify that you are referring to a "transport translator" that could need codec-specific knowledge to do the packetization from frame/stream to datagram transport. 

What happens if a translator receives a RESET_STREAM frame?  Is it still required to do a graceful close?  Maybe just require the translator to echo the behavior that it sees (e.g. if it receives a graceful close, then send one, if it receives a RESET_STREAM, then send that along, unless it is forwarding datagrams).

mengelbart:
Thanks!

I added another commit with the changes you suggested. I added a sentence about forwarding RESET_STREAM frames to section 6.1 because I think it might be confusing to reference the concept of closing streams via RESET_STREAM frames early in the document (even before the mapping to streams is introduced).
Alternatively, we could move the topology considerations section to the back of the document.

aboba:
"Once the RTP packet has been completely transmitted, the sender MUST gracefully close the stream."

[BA] How does the translator know that the RTP packet has been completely transmitted?  It could know if it receives a FIN, or alternatively if there is a length field at the beginning of the packet. But what happens if it never receives the full length (or a RESET_STREAM). 

It would reduce latency if the translator could implement "cut through" behavior, rather than "store and forward", sending what it receives (including a FIN or a RESET_STREAM). In that case, is the above text necessary?

mengelbart:
I am not sure I understand your question. I added the sentence in https://github.com/mengelbart/rtp-over-quic-draft/pull/17 to make sure a sender closes the stream after each packet so that a receiver knows when the packet was completely received. We may delete it again if we decide to add a length field, but I'd like to keep it for now. If a translator never receives the full length there will be a RESET_STREAM, which it then has to forward:

> A translators that translates between two endpoints, which are both connected via QUIC, MUST forward RESET\_STREAM frames received from one end to the other end, unless it is forwarding the RTP packets on QUIC datagrams.

PR #57, opened 2023-01-13
Generalize which packets should be included

mengelbart:
RTP packets can now also be sent in streams, thus it makes sense to say RTP packet (which could be in a datagram or stream) instead of datagram frame here.


mengelbart:
Will be solved in #60

PR #60, opened 2023-01-16
Address many/most of comments from @goelvidhi

### LINKED ISSUE #53: Review of RTP over QUIC draft (Vidhi Goel)
Posted to the AVTCORE WG mailing list: https://mailarchive.ietf.org/arch/msg/avt/BA7LGvYhodx-PnnaKKPj2gcnzBc/

Hello Authors,

I have read and reviewed draft-ietf-avtcore-rtp-over-quic-01 and below are my comments.

General comment - I think this draft will be really helpful for developers of real time applications in migrating to QUIC (from UDP and DTLS). The current draft touches on some important aspects such as multiplexing, de-depulication of RTCP feedback fields etc. But it needs some more details and clarifications regarding sections like congestion control, connection migration, multi path capability.

Specific comments - I am providing my comments section wise and there are a mix of minor nits, typos and major comments.

Section 1. - A very big advantage of migrating to QUIC would be connection migration and MPQUIC. I think this should be mentioned in the intro and later described the impact of each of these on RTP. 
Section 3. - Current text - Such new media transport protocols may be covered elsewhere, e.g., in the MOQ WG. I don’t think reference to MOQ is needed here. 
Section 6. - Multiplexing RTP/RTCP and non RTP (eg. HTTP) - since this will be discussed on Dec 15th interim, I will leave out my comment until a decision has been made.
Section 6.1 - Current text - If it is known to either the sender, that a packet, which was not yet successfully and completely transmitted, is no longer needed.... Typo in the first part where “either” is not needed.
Section 7. - Typo in the word “additional” -  QUIC layer to the application instead of exchanging addtional
Section 7.1 - The list of replaceable RTCP reports appears without any pretext before “Receiver Reports”. A simple line that says what is coming next would solve it.
                   - Under Receiver reports->Fractions lost, current text - Later packets SHOULD be ignored, since they may still be in flight, unless other QUIC packets that were sent after the datagram frame,. Why only sent after the datagram frame (and not stream) - maybe reframe the sentence to say RTP packet instead?
		   - Under Receiver reports->Highest Sequence Number received - This field is not clear. It would be clearer if it’d say what does this field mean in RTP and can QUIC provide the exact same feedback from its ACKs.
		   - Negative acknowledgements - Probably make the recommendation clearer that since QUIC can provide all the information that a negative ack provides, there is no need to use RTCP negative Ack.
   		   - ECN feedback - same comment as above, clearly state that reporting of ECN feedback should be done via QUIC instead of RTCP feedback.
Section 7.2 - English nit in current text - A QUIC receiver can also not calculate. Suggestion, Nor can a QUIC receiver calculate..
Section 8 - This section needs some work IMO with regards to,
	1. What spec should one comply to for a real time congestion controller? (I don’t think RFC 9002 is the right compliance for real time CCs)
	2. Is bandwidth estimation the only thing needed by media codecs or does QUIC CC need to do more than that?
	3. If a real time CC can’t comply to 1., does it need to use both real time CC as well as QUIC CC?

Section 10.2 - This is a sub section for impact of connection migration and it would be good to add one for MP QUIC as Multipath offers more benefits than migration. And, both the sections should clearly state if the RTP layer needs to do anything on path change.
		     - This line caught my attention - Application layer congestion control mechanisms (and also packet repair schemes such as retransmissions) need to be prepared to cope with such spikes.  Isn’t this something they already deal with UDP anyway? I wasn’t sure the reason for mentioning this explicitly.

Common to the entire draft - I think the word “unreliable” before datagrams doesn’t really add to the value of the draft as we all know datagrams are not retransmitted and the word “unreliable” doesn’t bring anything good to the table. Perhaps that can be removed at some places. The word congestion control is well known in the community but I think most real time drafts use the word “rate adaptation algorithm” and that word makes more sense to me than congestion control. Happy to hear what the authors (and others) think.

I can send out a PR for some of these comments.

Thanks,
Vidhi

SpencerDawkins:
This should be worth reviewing now, while I work on larger topics. 

Comments and corrections are graciously welcomed. 

fixes #53

### DIFF
--- a/draft-ietf-avtcore-rtp-over-quic.md
+++ b/draft-ietf-avtcore-rtp-over-quic.md
@@ -30,7 +30,7 @@ author:
 This document specifies a minimal mapping for encapsulating RTP and RTCP packets
 within QUIC. It also discusses how to leverage state from the QUIC
 implementation in the endpoints to reduce the exchange of RTCP packets and how
-to implement congestion control.
+to implement rate adaptation and congestion control.
 
 --- middle
 
@@ -76,7 +76,7 @@ when, and only when, they appear in all capitals, as shown here.
 The following terms are used:
 
 Datagram:
-: Datagrams exist in UDP as well as in QUICs unreliable datagram extension. If not explicitly noted
+: Datagrams exist in UDP as well as in QUIC's unreliable datagram extension. If not explicitly noted
 differently, the term datagram in this document refers to a QUIC Datagram as defined in
 {{!RFC9221}}.
 
@@ -113,7 +113,7 @@ On top of QUIC, this document defines an encapsulation of RTP and RTCP packets.
 
 The scope of this document is limited to carrying RTP over QUIC. It does not attempt
 to enhance QUIC for real-time media or define a replacement or evolution of RTP.
-Such new media transport protocols may be covered elsewhere, e.g., in the MOQ WG.
+Work to map other media transport protocols is under way elsewhere in the IETF.
 
 Protocols for negotiating connection setup and the associated parameters are defined
 separately, e.g., in {{?I-D.draft-dawkins-avtcore-sdp-rtp-quic}}.
@@ -122,8 +122,8 @@ separately, e.g., in {{?I-D.draft-dawkins-avtcore-sdp-rtp-quic}}.
 
 This document introduces a mapping of the Real-time Transport Protocol (RTP) to
 the QUIC transport protocol. RTP over QUIC allows the use of QUIC streams and
-unreliable QUIC datagrams to transport real-time data, and thus, the QUIC
-implementation MUST support QUICs unreliable datagram extension, if RTP packets
+QUIC datagrams to transport real-time data, and thus, the QUIC
+implementation MUST support QUIC's datagram extension, if RTP packets
 should be sent over QUIC datagrams. Since datagram frames cannot be fragmented,
 the QUIC implementation MUST also provide a way to query the maximum datagram
 size so that an application can create RTP packets that always fit into a QUIC
@@ -140,16 +140,16 @@ in {{Section 2.1 of ?I-D.draft-ietf-masque-h3-datagram}}. RTP over QUIC uses a
 flow identifier to replace the network address and port number to multiplex many
 RTP sessions over the same QUIC connection.
 
-A congestion controller can be plugged in to adapt the media bitrate to the
-available bandwidth. This document does not mandate any congestion control
+A rate adaptation algorithm can be plugged in to adapt the media bitrate to the
+available bandwidth. This document does not mandate any specific rate adaptation
 algorithm. Some examples include Network-Assisted Dynamic Adaptation (NADA)
 {{!RFC8698}} and Self-Clocked Rate Adaptation for Multimedia (SCReAM)
-{{!RFC8298}}. These congestion control algorithms require some feedback about
+{{!RFC8298}}. These rate adaptation algorithms require some feedback about
 the network's performance to calculate target bitrates. Traditionally this
 feedback is generated at the receiver and sent back to the sender via RTCP.
 Since QUIC also collects some metrics about the network's performance, these
 metrics can be used to generate the required feedback at the sender-side and
-provide it to the congestion controller to avoid the additional overhead of the
+provide it to the rate adaptation algorithm to avoid the additional overhead of the
 RTCP stream.
 
 ## Supported RTP Topologies {#topologies}
@@ -215,8 +215,8 @@ the string "-" and an experiment name to the identifier.
 
 This section describes the encapsulation of RTP/RTCP packets in QUIC.
 
-QUIC supports two transport methods: reliable streams {{!RFC9000}} and
-unreliable datagrams {{!RFC9221}}. This document specifies mappings of RTP to
+QUIC supports two transport methods: streams {{!RFC9000}} and
+datagrams {{!RFC9221}}. This document specifies mappings of RTP to
 both of the transport modes. Senders MAY combine both modes by sending some
 RTP/RTCP packets over the same or different QUIC streams and others in QUIC
 datagrams.
@@ -399,7 +399,7 @@ In general, applications MAY send RTCP without any restrictions. This document
 specifies a baseline for replacing some of the RTCP packet types by mapping the
 contents to QUIC connection statistics. Future documents can extend this mapping
 for other RTCP format types. It is RECOMMENDED to expose relevant information
-from the QUIC layer to the application instead of exchanging addtional RTCP
+from the QUIC layer to the application instead of exchanging addtiaddtionalional RTCP
 packets, where applicable.
 
 This section discusses what information can be exposed from the QUIC connection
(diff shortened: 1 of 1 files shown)


draft-ietf-avtcore-rtp-over-quic.md
```diff
@@ -371,7 +371,7 @@ In general, applications MAY send RTCP without any restrictions. This document
 specifies a baseline for replacing some of the RTCP packet types by mapping the
 contents to QUIC connection statistics. Future documents can extend this mapping
 for other RTCP format types. It is RECOMMENDED to expose relevant information
-from the QUIC layer to the application instead of exchanging addtional RTCP
+from the QUIC layer to the application instead of exchanging addtiaddtionalional RTCP
```
mengelbart, line 374: additional

draft-ietf-avtcore-rtp-over-quic.md
```diff
@@ -411,42 +411,45 @@ extension, that can improve the precision of the feedback from QUIC is
...
+  * *Delay since last SR*: This field is not required when the receiver reports
     are entirely replaced by QUIC feedback.
 * *Negative Acknowledgments* (`PT=205`, `FMT=1`, `Name=Generic NACK`,
   {{!RFC4585}})
   * The generic negative acknowledgment packet contains information about
     packets which the receiver considered lost. {{Section 6.2.1. of !RFC4585}}
     recommends to use this feature only, if the underlying protocol cannot
     provide similar feedback. QUIC does not provide negative acknowledgments,
-    but can detect lost packets through acknowledgments.
+    but can detect lost packets based on the Gap numbers contained in QUIC ACK frames.
```
mengelbart, line 441: Should we also reference [Section 6 of RFC 9002](https://www.rfc-editor.org/rfc/rfc9002.html#section-6) here?
```suggestion
    but can detect lost packets based on the Gap numbers contained in QUIC ACK frames {{Section 6 of !RFC9002}}.
```

draft-ietf-avtcore-rtp-over-quic.md
```diff
@@ -491,22 +494,22 @@ congestion control to avoid overloading the network.
 
 QUIC is a congestion controlled transport protocol. Senders are required to
 employ some form of congestion control. The default congestion control specified
-for QUIC is an alogrithm similar to TCP NewReno, but senders are free to choose
+for QUIC is an alogrithm similar to TCP NewReno {{RFC9002}}, but senders are free to choose
```
mengelbart, line 497:
```suggestion
for QUIC is an alogrithm similar to TCP NewReno {{!RFC9002}}, but senders are free to choose
```
`!` makes it a normative reference, `?` for informative.

draft-ietf-avtcore-rtp-over-quic.md
```diff
@@ -411,42 +411,45 @@ extension, that can improve the precision of the feedback from QUIC is
...
+The list of RTCP Receiver Reports that could be replaced by feedback from QUIC follows:
+
 * *Receiver Reports* (`PT=201`, `Name=RR`, {{!RFC3550}})
-  * *Fraction lost*: The fraction of lost packets can be directly infered from
+  * *Fraction lost*: When RTP packets are carried in QUIC datagrams, the fraction of lost packets can be directly inferred from
     QUIC's acknowledgments. The calculation SHOULD include all packets up to the
     acknowledged RTP packet with the highest RTP sequence number. Later packets
     SHOULD be ignored, since they may still be in flight, unless other QUIC
-    packets that were sent after the datagram frame, were already acknowledged.
+    packets that were sent after the RTP packet frame, were already acknowledged.
```
mengelbart, line 421:
```suggestion
    packets that were sent after the RTP packet, were already acknowledged.
```

In streams, the RTP packet can be spread over many frames.

draft-ietf-avtcore-rtp-over-quic.md
```diff
@@ -411,42 +411,45 @@ extension, that can improve the precision of the feedback from QUIC is
...
   * *Cumulative lost*: Similar to the fraction of lost packets, the cumulative
-    loss can be infered from QUIC's acknowledgments including all packets up to
+    loss can be inferred from QUIC's acknowledgments including all packets up to
     the latest acknowledged packet.
   * *Highest Sequence Number received*: The highest sequence number received is
-    the sequence number of all RTP packets that were acknowledged.
-  * Interarrival jitter: If QUIC acknowledgments carry timestamps as described
+    the highest sequence number of all RTP packets carried in a QUIC packet that was acknowledged.
```
mengelbart, line 426: In another PR I added some more text to explain what the highest sequence number is in RTP:
```
  * Highest Sequence Number received*: In RTCP, this field is a 32-bit field
    that contains the highest sequence number a receiver received in an RTP
    packet and the count of sequence number cycles the receiver has observed. A
    sender sends RTP packets in QUIC packets and receives acknowledgments for
    the QUIC packets. By keeping a mapping from a QUIC packet to the RTP packets
    encapsulated in that QUIC packet, the sender can infer the highest sequence
    number and number of cycles seen by the receiver from QUIC acknowledgments.
```

Does it make sense to add this explanation, and if yes, should we add it to this PR?

goelvidhi: I think we should add this explanation. (Whether it is done here or in a separate PR, authors can decide that.)

draft-ietf-avtcore-rtp-over-quic.md
goelvidhi: This reference and text is confusing. The text says TCP Reno but then it refers to QUIC Reno. I would just change this to, `The default congestion control specified for QUIC is Reno {{!RFC 9002}}`

draft-ietf-avtcore-rtp-over-quic.md
```diff
@@ -491,22 +494,22 @@ congestion control to avoid overloading the network.
...
 employ some form of congestion control. The default congestion control specified
-for QUIC is an alogrithm similar to TCP NewReno, but senders are free to choose
+for QUIC is an alogrithm similar to TCP NewReno {{RFC9002}}, but senders are free to choose
 any congestion control algorithm as long as they follow the guidelines specified
 in {{Section 3 of ?RFC8085}}.
 
-RTP does not specify a congestion controller, but provides feedback formats for
-congestion control (e.g. {{!RFC8888}}) as well as different congestion control
+RTP does not specify a rate adaptation algorithm, but provides feedback formats for
+congestion control (e.g. {{!RFC8888}}) as well as different rate adaptation
```
goelvidhi, line 502: replace congestion control with rate adaptation?

draft-ietf-avtcore-rtp-over-quic.md
SpencerDawkins: Corrected.

draft-ietf-avtcore-rtp-over-quic.md
SpencerDawkins: @goelvidhi is right - this text/reference **is** scrambled. My thoughts are

- TCP NewReno is [RFC6582]
- the default QUIC congestion controller is [RFC9002] 
- [RFC9002] says 
> This document specifies a sender-side congestion controller for QUIC similar to TCP NewReno [RFC6582].

So, putting all this together, I get 

> The default congestion control specified for QUIC in {{RFC9002}} is similar to TCP NewReno {{RFC6582}}

which (IIUC) gives the references for QUIC's default congestion controller and TCP NewReno correctly. 

Please let me know if this is a Really Bad Idea.

draft-ietf-avtcore-rtp-over-quic.md
SpencerDawkins: Fixed.

draft-ietf-avtcore-rtp-over-quic.md
SpencerDawkins: @mengelbart, I agree with @goelvidhi about adding this text. I put it in #60, so if it's also in another PR (which I haven't looked for yet), that should also be fine.

draft-ietf-avtcore-rtp-over-quic.md
SpencerDawkins: This is, of course, going to be maximally confusing to the reader, because RFC 8888 **says** it's enabling congestion control, but I think this is a good change to make for consistency in RTP-over-QUIC. 

Looking more closely at this text, RFC 8888 isn't technically part of "RTP", because it doesn't update any other RFC, so I think this is more accurate: 

> RTP itself does not specify a congestion control algorithm, but {{!RFC8888}} defines an RTCP
> feedback message intended to enable rate adaptation for interactive real-time traffic using RTP,
> and successful rate adaptation will accomoplish congestion control as well.
> Various rate adaptation algorithms for real-time media are defined in separate RFCs
> (e.g. SCReAM {{!RFC8298}} and NADA {{!RFC8698}}).

And, looking more closely at the Abstract and Introduction, I think I should be naming both congestion control and rate adaptation in both places, and then favoring the use of rate adaptation thereafter. 

Please let me know if this seems wrong or unhelpful.

SpencerDawkins:
@mengelbart - I have a couple of open questions in the comments on this PR for you - do you think this PR is ready to commit, so we can focus on more specific issues (some of which I've opened in the last day or two)?

draft-ietf-avtcore-rtp-over-quic.md
mengelbart: Sounds good to me.

draft-ietf-avtcore-rtp-over-quic.md
mengelbart: Sounds mostly fine to me, but I wonder if we need the part "and successful rate adaptation will accomoplish congestion control as well."? I think it might bring up the question of how exactly we define both and what the difference is.

mengelbart: Oh, I just saw #61 is about this question. If we have some text early in the document that defines both terms, the question might have been answered already.

mengelbart:
Thanks, @SpencerDawkins! I think I replied to all questions, and I think we can discuss the congestion control vs. rate adaptation question in #61, and we can merge this PR.

Issue #62, opened 2023-01-20
QUIC congestion control, real-time congestion control, and (potentially) both, for the same QUIC connection


SpencerDawkins:
@goelvidhi had several questions about this interaction in her document review (sent to the mailing list, entered into #53 by @aboba), and it's worth mentioning this in a separate issue. 

- What spec should one comply to for a real time congestion controller? (I don’t think RFC 9002 is the right compliance for real time CCs)
- Is bandwidth estimation the only thing needed by media codecs or does QUIC CC need to do more than that?
- If a real time CC can’t comply to 1., does it need to use both real time CC as well as QUIC CC?

This is related to, but not the same as, #59 (if you don't want your media treated like H3 payloads, how do you tell the other endpoint that it's real-time media?).

mengelbart:
> * What spec should one comply to for a real time congestion controller? (I don’t think RFC 9002 is the right compliance for real time CCs)

Since we are running RTP on top of QUIC and one of our goals is not to change or extend QUIC if possible, I would argue that RFC 9002 is at least one of the specs that a congestion controller should comply with. RFC 9002 says:

> A sender can unilaterally choose a different algorithm to use, such as CUBIC [[RFC8312](https://www.rfc-editor.org/rfc/rfc9002.html#RFC8312)].
>
> If a sender uses a different controller than that specified in this document, the chosen controller MUST conform to the congestion control guidelines specified in [Section 3.1](https://www.rfc-editor.org/rfc/rfc8085#section-3.1) of [[RFC8085](https://www.rfc-editor.org/rfc/rfc9002.html#RFC8085)].

which is reflected in the current draft in section 8. The section references RTP congestion control mechanisms but doesn't define any requirements. I would suggest adding [RFC 8836 (Congestion Control Requirements for Interactive Real-Time Media)](https://datatracker.ietf.org/doc/rfc8836/) here.
 
> * Is bandwidth estimation the only thing needed by media codecs or does QUIC CC need to do more than that?

I don't think there is anything that a media codec needs from a congestion controller. Calculating a bandwidth estimation helps because it allows the codec to produce media at a rate close to the maximum rate at which data can be transmitted to the receiver in real time. There may be more detailed signals available to improve the experience at the receiver, e.g., by changing the resolution or generating a new key fream, but I don't think this is the responsibility of congestion control.


> * If a real time CC can’t comply to 1., does it need to use both real time CC as well as QUIC CC?

[...]

Issue #59, opened 2023-01-16
CC/RA: Explain how to avoid nested congestion controller loops for real time media


SpencerDawkins:
The current draft says (in "Congestion Control at the Application Layer"), 

> If the application implements full congestion control rather than just a
> bandwidth estimation at the application layer using a congestion controller that
> satisfies the requirements of {{Section 7 of !RFC9002}}, and the connection is
> only used to send real-time media which is subject to the application layer
> congestion control, **it is RECOMMENDED to disable any other congestion control
> that is possibly running at the QUIC layer**. Disabling the additional congestion
> controllers helps to avoid any interference between the different congestion
> controllers.

We have talked about this for at least a couple of years, back when "RTP over QUIC" was still in the QUIC working group, but as I understand it, there's no way to do this now (in a standardized way), and a plain reading of the emphasized text makes me think what's being described would require a QUIC extension. 

I don't care how we actually do this, but we need to write down how to do it!

goelvidhi:
Just noting down the points from my email that needs addressing,

```
Section 8 - This section needs some work IMO with regards to,
1. What spec should one comply to for a real time congestion controller? (I don’t think RFC 9002 is the right compliance for real time CCs)
2. Is bandwidth estimation the only thing needed by media codecs or does QUIC CC need to do more than that?
3. If a real time CC can’t comply to 1., does it need to use both real time CC as well as QUIC CC?
```

SpencerDawkins:
From discussion with @joerg-ott and @mengelbart -

- this is mostly about how a sending application tells a QUIC implementation what general type of congestion controller the QUIC implementation should use (say, NADA vs NewReno, "please DON'T use an h3-optimized congestion controller for media")
- are we expecting the QUIC implementation to figure this out automagically? 
- how badly are h3-optimized congestion controllers going to work for application-limited, interactive applications?
- Mathis work shows that rate-limited RTP rarely encounters QUIC limits that bulk transfer traffic would encounter

SpencerDawkins:
This issue is about the proposal to disable QUIC's congestion controller, since a rate-limiting congestion controller is already in use at the application level.

mengelbart:
I replied to @goelvidhi s comments in #62 .

I agree that the quoted text above is a bit unclear and maybe mixes a few things together that should be separated.

[...]

PR #64, opened 2023-02-13
First draft of definitions for congestion control and rate adaptation

### LINKED ISSUE #61: Distinguish clearly between "congestion control" and "rate adaptation"
@goelvidhi mentioned this in her document review (sent to the mailing list, entered into #53 by @aboba), and I worked on this in #60, but where I ended up, was 

- it's helpful to distinguish between congestion control in general, and rate adaptation for interactive low-latency media, **but**
- a significant number of RTP-centric specifications (for example, [RFC 8888](https://datatracker.ietf.org/doc/html/rfc8888)) use the term "congestion control" to describe what we're calling rate adaptation in this specification. 

I note that [RFC 8298](https://datatracker.ietf.org/doc/html/rfc8298) distinguishes between 

- Network Congestion Control, 
- Sender Transmission Control, and 
- Media Rate Control. 

Something like that, early in this document, may be helpful.

SpencerDawkins:
This is a starting place for several related issues (#61, #62, and #59). 

Closes #61

### DIFF
--- a/draft-ietf-avtcore-rtp-over-quic.md
+++ b/draft-ietf-avtcore-rtp-over-quic.md
@@ -75,6 +75,9 @@ when, and only when, they appear in all capitals, as shown here.
 
 The following terms are used:
 
+Congestion Control:
+: A mechanism to limit the aggregate amount of data that has been sent over a path to a receiver, but has not been acknowledged by the receiver. This prevents a sender from overwhelming the capacity of a path between a sender and a receiver, causing some outstanding data to be discarded before it can be received by the receiver and acknowledged to the sender.
+
 Datagram:
 : Datagrams exist in UDP as well as in QUIC's unreliable datagram extension. If not explicitly noted
 differently, the term datagram in this document refers to a QUIC Datagram as defined in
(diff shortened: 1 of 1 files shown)


joerg-ott:
The above definitions state that rate adaptation is "distinct" from congestion control. But the distinction mentioned only covers the case in which one choose across multiple media flows and some aggregate congestion control can be performed. This appears to be a special case. Rate adaptation is generally one way to achieve congestion control, it would seem, and maybe in some cases the only one. Needs more thinking.

SpencerDawkins:
> The above definitions state that rate adaptation is "distinct" from congestion control. But the distinction mentioned only covers the case in which one choose across multiple media flows and some aggregate congestion control can be performed. This appears to be a special case. Rate adaptation is generally one way to achieve congestion control, it would seem, and maybe in some cases the only one. Needs more thinking.

@joerg-ott, I agree. I've rephrased this in my latest commit. 

We will definitely need more thinking here - I see that Christian @Huitema has a [specific proposal for congestion response in MOQ](https://mailarchive.ietf.org/arch/msg/moq/sWDaR6lRu3tBLOvHkth-sipbf4Q/). That may not be directly applicable to RTP-over-QUIC, but it's a good example of the kind of this we should also consider in your draft. 

But given that we have other issues for adding specific descriptions of rate control and this PR only targeted adding definitions, I think this PR is ready to merge.

Issue #45, opened 2022-10-24
Receiver can't request stream cancellation vie STOP_SENDING


mengelbart:
In #43 we added a length field to RTP packets that are sent over QUIC streams and allowed multiple packets per stream. Until now, it was possible for a receiver to request cancellation of a stream, if a packet is not needed anymore, e.g., because a deadline has passed. 

Now, since the receiver does not know how many packets to expect on a stream, it cannot know if it is safe to request cancellation, because the following packets may still be relevant.

One possible solution is to go back to one Application Data Unit (ADU) (e.g., one media frame) per stream, while still allowing this ADU to be fragmented into multiple RTP packets.

aboba:
If SVC layers are sent on different streams, a receiver can send a STOP_SENDING in order to unsubscribe from a layer.  The receiver can do this if it is encountering too much loss.  If the receiver unsubscribes from a layer (and presumably, any higher layers), then it would have lost interest in future packets on those layers (at least until there is sufficient bandwidth to receive them).

mengelbart:
That is an interesting feature that I didn't think about before. Does it require some signaling to tell the receiver what layers are sent on which stream? How would a sender know when to restart sending the higher layer(s)?

When we discussed this problem in London, there was a tendency to require the sender to close a stream after exactly one ADU. But it is unclear what an ADU is in different payload formats. For some audio codecs, that could lead to streams being closed after each frame containing only a few ms of data. That would then lead to a very large number of new streams. Too many streams could be a problem itself (see also #49), but it may also be unnecessary to transmit each of these audio frames independently.

aboba:
The receiver would need to know that the layers were segregated by stream. This could occur because the receiver requested a layer or because the sender set it up that way. If the receiver is getting one-way delay info, it could ask for more layers if the delay was stable and close to the transmission line, indicating low queuing and loss. But without probing, neither sender nor receiver can estimate the bottleneck bandwidth if the rate is well below that. So requesting more layers could result in increasing delays, indicating queue buildup.

mengelbart:
[...]

PR #88, opened 2023-05-12
Add considerations for STOP\_SENDING

mengelbart:


mengelbart:
closes #45

SpencerDawkins (commented):
Thank you for working through the comments about this. 

I think (but could be wrong) that we are complicating our work by saying "this is how ROQ works", and then having to clarify that "this is a QUIC streams thing (like STOP\_SENDING), and the rules are different if you're using QUIC datagrams (as below), and (of course) it's legal for ROQ endpoints to use both". 

But that's a question for another issue. On **this issue** ... 

> QUIC allows an application to abort reading the stream and specify an error code
> {{Section 3.5 of !RFC9000}}. The QUIC transport can signal this to the sending
> side by sending a STOP\_SENDING frame. The sender SHOULD not interpret the
> reception of STOP\_SENDING as an indication that the receiver lost interest in
> the media stream as a whole, but rather that a part of the media stream was not
> received timely and some recovery is necessary. 

Hmmm. Is this a normative requirement on QUIC, beyond RFC 9000/RFC 9221? I think we might express this better as 

The **sending RTP application** SHOULD not interpret the reception of STOP\_SENDING as an indication that the receiver lost interest in the media stream as a whole, but rather that a part of the media stream was not received timely and some recovery **is probably** necessary.

> The recovery may be supported by
> dropping some parts of the media stream entirely and restarting to send new
> media frames on new QUIC streams. While STOP\_SENDING includes an error code, it
> does not allow the receiving side to inform the sender where exactly it wishes
> to restart receiving media. A sender that receives STOP\_SENDING MUST NOT
> transmit any frames already sent on the QUIC stream, on which STOP\_SENDING was
> received, on a new QUIC stream. 

I was confused here. Is this saying the (sending RTP application) MUST NOT retransmit **media** frames on the new stream, or MUST NOT retransmit QUIC frames on the new stream?

> A sender SHOULD instead continue to send media
> frames on new QUIC streams starting with the first frame that was not
> transmitted on the stream that received STOP\_SENDING.

Skipping down to 

> **Editor's note:** A receiver cannot cancel a certain frame but still receive
> retransmissions for a frame the was following on the same stream using
> STOP\_SENDING, because STOP\_SENDING does not include an offset which would
> allow signaling where retransmissions should continue. If this is a required
> feature for RoQ, it could be implemented using an extended STOP\_SENDING frame
> as, for example, proposed in {{?I-D.draft-thomson-quic-enough-00}}

Do we have a complete list of QUIC extension RFCs and drafts we think might be useful in ROQ? That would be a fine thing to add.

mengelbart:
> The sending RTP application SHOULD not interpret the reception of STOP_SENDING as an indication that the receiver lost interest in the media stream as a whole, but rather that a part of the media stream was not received timely and some recovery is probably necessary.

Sounds good. I will change it.

> I was confused here. Is this saying the (sending RTP application) MUST NOT retransmit media frames on the new stream, or MUST NOT retransmit QUIC frames on the new stream?

This should be media frames, I will add it to clarify. 

> Do we have a complete list of QUIC extension RFCs and drafts we think might be useful in ROQ? That would be a fine thing to add.

That is a good idea, I opened #92

mengelbart:
I am not sure I understand your comment here:

> I think (but could be wrong) that we are complicating our work by saying "this is how ROQ works", and then having to clarify that "this is a QUIC streams thing (like STOP_SENDING), and the rules are different if you're using QUIC datagrams (as below), and (of course) it's legal for ROQ endpoints to use both".
> 
> But that's a question for another issue. On this issue ...

Would you mind opening that issue?

SpencerDawkins:
Preferring "RTP sender", rather than "sender".

SpencerDawkins:
> I am not sure I understand your comment here:
> 
>> I think (but could be wrong) that we are complicating our work by saying "this is how ROQ works", and then having to clarify that "this is a QUIC streams thing (like STOP_SENDING), and the rules are different if you're using QUIC datagrams (as below), and (of course) it's legal for ROQ endpoints to use both".
>>
>> But that's a question for another issue. On this issue ...
>
> Would you mind opening that issue?

Let's wait to think about this some more, but I'll open an issue for it when it becomes an issue. :grinning:

PR #91, opened 2023-05-16
Define error codes

mengelbart:
Actual codes are still missing to avoid redefining them multiple times if we add more codes.

### DIFF
--- a/draft-ietf-avtcore-rtp-over-quic.md
+++ b/draft-ietf-avtcore-rtp-over-quic.md
@@ -1197,19 +1197,15 @@ ROQ\_NO\_ERROR (0x????):
 : No error. This is used when the connection or stream needs to be closed, but
 there is no error to signal.
 
-ROQ\_GENERAL\_PROTOCOL\_ERROR (0x????):
-: Peer violated protocol requirements in a way that does not match a more
-specific error code or endpoint declines to use the more specific error code.
-
 ROQ\_INTERNAL\_ERROR (0x????):
 : An internal error has occured in the RoQ stack.
-
-ROQ\_STREAM\_CREATION\_ERROR (0x????):
-: The endpoint detected that its peer created a stream that it will not accept.
 
 ROQ\_PACKET\_ERROR (0x????):
 : Invalid payload format, e.g., length does not match packet, invalid flow id
 encoding, non-RTP on RTP-flow ID, etc.
+
+ROQ\_STREAM\_CREATION\_ERROR (0x????):
+: The endpoint detected that its peer created a stream that it will not accept.
 
 ROQ\_FRAME\_CANCELLED (0x????):
 : A receiving endpoint stopped reading a frame from a stream and requests new
--- a/draft-ietf-avtcore-rtp-over-quic.md
+++ b/draft-ietf-avtcore-rtp-over-quic.md
@@ -683,9 +683,10 @@ flow identifier of the RTP session in which the retransmission happens.
 
 # Connection Shutdown
 
-Both peers MAY close the connection for variety of reasons. If one of the peers
-wants to close the RoQ connection, the peer can use a QUIC CONNECTION\_CLOSE
-frame with one of the error codes defined in {{error-handling}}.
+Either peers MAY close the connection for variety of reasons. If one of the
+peers wants to close the RoQ connection, the peer can use a QUIC
+CONNECTION\_CLOSE frame with one of the error codes defined in
+{{error-handling}}.
 
 # Congestion Control and Rate Adaptation {#congestion-control}
 
@@ -1205,17 +1206,18 @@ ROQ\_PACKET\_ERROR (0x????):
 encoding, non-RTP on RTP-flow ID, etc.
 
 ROQ\_STREAM\_CREATION\_ERROR (0x????):
-: The endpoint detected that its peer created a stream that it will not accept.
+: The endpoint detected that its peer created a stream that violates the ROQ protocol.
 
 ROQ\_FRAME\_CANCELLED (0x????):
-: A receiving endpoint stopped reading a frame from a stream and requests new
-frames be sent on new streams using STOP\_SENDING or a sender notifies a
-receiver that retransmissions of a frame were stopped using RESET\_STREAM and
-new frames will be sent on new streams.
+: A receiving endpoint is using STOP_SENDING on the current stream to request
+new frames be sent on new streams. Similarly, a sender notifies a receiver that
+retransmissions of a frame were stopped using RESET\_STREAM and new frames will
+be sent on new streams.
 
 ROQ\_UNKNOWN\_FLOW\_ID (0x????):
 : An endpoint was unable to handle a flow identifier, e.g., because it was not
-signalled or it does not support multiplexing using arbitrary flow identifiers.
+signalled or because the endpoint does not support multiplexing using arbitrary
+flow identifiers.
 
 ROQ\_SIGNALING\_ERROR (0x????):
 : Parameters that were set during out of band signaling are incompatible with
--- a/draft-ietf-avtcore-rtp-over-quic.md
+++ b/draft-ietf-avtcore-rtp-over-quic.md
@@ -1364,6 +1364,20 @@ The "rtp-mux-quic" string identifies RoQ:
   Specification:
   : This document
 
+## RoQ Error Codes
+
+This document establishes a registry for RoQ error codes.
+
+| Name                          | Value  | Description                            | Specification  |
+| ----------------------------- | ------ | -------------------------------------- | -------------- |
+| ROQ\_NO\_ERROR                | 0x???? | No Error                               | TODO: This doc |
+| ROQ\_INTERNAL\_ERROR          | 0x???? | Internal Error                         | TODO: This doc |
+| ROQ\_PACKET\_ERROR            | 0x???? | Invalid payload format                 | TODO: This doc |
+| ROQ\_STREAM\_CREATION\_ERROR  | 0x???? | Invalid stream type                    | TODO: This doc |
+| ROQ\_FRAME\_CANCELLED         | 0x???? | Frame cancelled                        | TODO: This doc |
+| ROQ\_UNKNOWN\_FLOW\_ID        | 0x???? | Unknown Flow ID                        | TODO: This doc |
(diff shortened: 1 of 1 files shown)


SpencerDawkins:
This PR has a starting point - next steps are 

- present to WG and ask for comments
- add error codes to relevant sections of the draft

mengelbart:
I added text to the rest of the draft to explain which error codes should be used. Some error codes are not yet mentioned in the text, and I am not sure if we really need them:

* ROQ_GENERAL_PROTOCOL_ERROR
* ROQ_INTERNAL_ERROR
* ROQ_PACKET_ERROR

It may be enough to have one of them, or it may be helpful to have the option to be more specific.

(The PR is now based on #98 and should only be merged after that PR is merged, too)

draft-ietf-avtcore-rtp-over-quic.md
```diff
@@ -563,11 +573,17 @@ stream credits it will have to provide to the media-sending peer.
...
-extension to QUIC described in {{!RFC9221}}. QUIC datagrams preserve frame
-boundaries. Thus, a single RTP packet can be mapped to a single QUIC datagram
-without additional framing. Senders SHOULD consider the header overhead
-associated with QUIC datagrams and ensure that the RTP/RTCP packets, including
-their payloads, flow identifier, QUIC, and IP headers, will fit into path MTU.
+extension to QUIC described in {{!RFC9221}}. QUIC datagrams can only be used if
+the use of the extension was successfully negotiated during the QUIC handshake.
+If the use of datagrams was signalled using a signalling protocol, but the
+datagram extension was not negotiated during the handshake, a peer MAY close the
+connection with the ROQ\_SIGNALING\_ERROR error code.
```
SpencerDawkins, line 580: Hmmm. RFC 9000 says this:

> An endpoint MUST treat the receipt of a frame of unknown type as a connection error of type FRAME_ENCODING_ERROR.

We just had a conversation about whether a "frame of known type defined for an extension that has not been negotiated" would be treated as "a frame of unknown type" or some other kind of error, but the high order bit is that I don't think the RTP receiver will ever see that datagram - the QUIC receiver will have already punted it as a connection error.

draft-ietf-avtcore-rtp-over-quic.md
```diff
@@ -610,6 +626,12 @@ application. RTP retransmissions can be done in the same RTP session or a
 separate RTP session {{!RFC4588}} and the flow identifier MUST be set to the
 flow identifier of the RTP session in which the retransmission happens.
 
+# Connection Shutdown
+
+Both peers MAY close the connection for variety of reasons. If one of the peers
```
SpencerDawkins, line 631: I think this is "Either peer MAY close the connection for a variety of reasons."

draft-ietf-avtcore-rtp-over-quic.md
```diff
@@ -1079,6 +1101,40 @@ following Payload-specific RTP Feedback (PSFB) feedback.
...
+ROQ\_PACKET\_ERROR (0x????):
+: Invalid payload format, e.g., length does not match packet, invalid flow id
+encoding, non-RTP on RTP-flow ID, etc.
+
+ROQ\_STREAM\_CREATION\_ERROR (0x????):
+: The endpoint detected that its peer created a stream that it will not accept.
+
+ROQ\_FRAME\_CANCELLED (0x????):
+: A receiving endpoint stopped reading a frame from a stream and requests new
+frames be sent on new streams using STOP\_SENDING or a sender notifies a
```
SpencerDawkins, line 1125: > : A receiving endpoint stopped reading a frame from a stream and requests new
> frames be sent on new streams using STOP\_SENDING 

I think that's technically correct, but if the following is also correct, it might be clearer. 

> : A receiving endpoint is using STOP\_SENDING on the current stream to request new
> frames be sent on new streams.

draft-ietf-avtcore-rtp-over-quic.md
```diff
@@ -1079,6 +1101,40 @@ following Payload-specific RTP Feedback (PSFB) feedback.
...
+receiver that retransmissions of a frame were stopped using RESET\_STREAM and
+new frames will be sent on new streams.
+
+ROQ\_UNKNOWN\_FLOW\_ID (0x????):
+: An endpoint was unable to handle a flow identifier, e.g., because it was not
+signalled or it does not support multiplexing using arbitrary flow identifiers.
```
SpencerDawkins, line 1131: Perhaps 

> : An endpoint was unable to handle a flow identifier, e.g., because it was not
> signalled or **because the endpoint** does not support multiplexing using arbitrary flow identifiers.

draft-ietf-avtcore-rtp-over-quic.md
```diff
@@ -1079,6 +1101,40 @@ following Payload-specific RTP Feedback (PSFB) feedback.
...
+
+ROQ\_SIGNALING\_ERROR (0x????):
+: Parameters that were set during out of band signaling are incompatible with
+the connection properties that were negotiated when the connection was
+established using transport parameters.
+
```
SpencerDawkins, line 1137: :exploding_head: I'm lost on this one. Does 

> the connection properties that were negotiated when the connection was
> established using transport parameters.

refer to QUIC transport parameters? If so, it would be good to say that, but we are using "parameters" to refer to ROC parameters, so I'm not sure what's being described. 

Please advise.

draft-ietf-avtcore-rtp-over-quic.md
```diff
@@ -1079,6 +1101,40 @@ following Payload-specific RTP Feedback (PSFB) feedback.
...
+ROQ\_INTERNAL\_ERROR (0x????):
+: An internal error has occured in the RoQ stack.
+
+ROQ\_PACKET\_ERROR (0x????):
+: Invalid payload format, e.g., length does not match packet, invalid flow id
+encoding, non-RTP on RTP-flow ID, etc.
+
+ROQ\_STREAM\_CREATION\_ERROR (0x????):
+: The endpoint detected that its peer created a stream that it will not accept.
```
SpencerDawkins, line 1121: The use of this error code in the document, in 

> To send RTP/RTCP packets over QUIC streams, a sender MUST open a
> new unidirectional QUIC stream. Streams are unidirectional because there is no
> synchronous relationship between sent and received RTP/RTCP packets. A peer that
> receives a bidirectional stream with a flow identifier that is associated with
> an RTP or RTCP stream, SHOULD stop reading from the stream and send a
> STOP\_SENDING frame with the application protocol error code set to
> ROQ\_STREAM\_CREATION\_ERROR.

describes this situation as a fundamental protocol violation. I read 

> ROQ\_STREAM\_CREATION\_ERROR (0x????):
> : The endpoint detected that its peer created a stream that **it will not accept.**

as more of a decision by the endpoint, rather than what the endpoint does to conform to the protocol specification. 

Maybe something like 

> ROQ\_STREAM\_CREATION\_ERROR (0x????):
> : The endpoint detected that its peer created a stream that **violates the ROQ protocol.**

SpencerDawkins (approved):
Nice work! I apologize for the number of comments ...

draft-ietf-avtcore-rtp-over-quic.md
```diff
@@ -510,6 +517,9 @@ A RoQ receiver that is no longer interested in reading a certain partition of
 the media stream MAY signal this to the sending peer using a STOP\_SENDING
 frame.
 
+In both cases, the error code of the RESET\_STREAM frame or the STOP\_SENDING
+frame MUST be set to ROQ\_FRAME\_CANCELLED.
+
 When a RoQ sender receives a STOP\_SENDING frame, the RoQ sender MUST open one
 or more new QUIC streams to send new media frames. Any media frame that has
```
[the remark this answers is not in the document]
mengelbart: That makes sense, but I will change this in #106 because it also changes this sentence.

draft-ietf-avtcore-rtp-over-quic.md
mengelbart: I think we are talking about two different error cases. One problem is a peer sending a datagram frame in a connection that did not negotiate the use of QUIC datagrams. In that case, I agree that the QUIC receiver will react first, and the RTP receiver will never know about it. The case I intended to cover is that the two peers used some out-of-band signaling before setting up the QUIC connection, and in that signaling, they agreed to use QUIC datagrams, but then during the QUIC handshake, one of the peers does not send the transport parameter to negotiate datagrams. In that QUIC connection, datagrams are not allowed, so the peer that was expecting to be allowed to use datagrams can immediately close the connection.

draft-ietf-avtcore-rtp-over-quic.md
mengelbart: Yes, it refers to QUIC transport parameters. This was meant for the case I mentioned in a comment above, where peers agree on using QUIC datagrams out-of-band, but then one of them does not allow datagrams during the QUIC handshake. In that case, the other peer might like to say: "Hey, we agreed on using datagrams, but now you don't allow me to use them so I just close the connection with a ROQ_SIGNALING_ERROR". Does it help to say "... established using QUIC transport parameters"?

draft-ietf-avtcore-rtp-over-quic.md
LPardue: @mengelbart's explanation makes sense to me

draft-ietf-avtcore-rtp-over-quic.md
LPardue: It might help to rename the error to be like "ROQ_REQUIREMENT_UNMET" or "ROQ_EXPECTATION_UNMET" that would more clearly articulate the intent. I'd then change the text to say something like

> Expectiations of the QUIC transport set by RoQ out-of-band signalling were not met by the QUIC connection.

LPardue (changes requested):
This is all going in the correct direction. However some things need tweaking before you merge:

* Define a new RoQ Error Code IANA registry e.g. like HTTP/3's https://www.iana.org/assignments/http3-parameters/http3-parameters.xhtml#http3-parameters-error-codes. Note that this will need to have registration procedures and you'll want to consider about greasing the code point space

* Pick real values for the error codes. This is harder when you're not done and those errors might need to change. One technique is to pick values in the upper code space and then prior to publication switch them to lower values. But that might risk more churn than is really worth it.

mengelbart:
@LPardue and I agreed on merging this PR first and then defining IANA procedures and real error codes in a follow-up PR.

PR #98, opened 2023-06-16
Clarify behaviour of STOP_SENDING and RESET_STREAM

mengelbart:
Second attempt to clarify the behaviour of STOP_SENDING and RESET_STREAM. I hope this is easier to understand than the current text and #96. It also adds subsections to add more structure to the whole subsection.


SpencerDawkins:
I sent some comments via email, but I think the current text is close enough to merge for IETF 117.

SpencerDawkins (approved):
I sent some comments via email, but I think the current text is close enough to merge for IETF 117.

PR #100, opened 2023-07-03
Add a list of considered header extensions

### LINKED ISSUE #95: Add header extensions to RTCP analysis
In #71 we added a section about header extensions to the RTCP analysis, but it does not include a table of all header extensions we considered. I think it makes sense to add one, just like we did for the RTCP packets, too.

mengelbart:
close #95


mengelbart:
close #95

PR #109, opened 2023-07-19
Add motivations section

### LINKED ISSUE #77: Add motivations for RoQ adoption
Add motivations for RQ work and adoption

Transport-level encryption with RQ.

Using QUIC as it evolves (connection migration, and especially multipath QUIC).

Thinking about policy aspects of connection migration between interfaces (cheap WiFi vs expensive cellular).

Potential connection sharing with other protocols over QUIC.

QUIC path MTU discovery, especially in multiparty environments (today conference server has no option except using the lowest path MTU for all conference participants). 

ACK piggybacking in QUIC if there are any QUIC frames going the other direction versus ACKs decreasing the available MTU for media. 

Being clear about how RQ congestion control and flow control work.

Possible advantage if audio is piggybacking on video in RQ - packaging multiple frames into a packet. 

Think about how multiple media quality variants would work - is there an advantage here?

Multiplexing real-time and non-real-time flows. 

Interaction between QUIC congestion control and media congestion control.

SpencerDawkins:
@mengelbart and @gchandok - I'm sure I'm leaving motivations out, so I'd appreciate your comments on this PR even more than usual!

closes #77

### DIFF
--- a/draft-ietf-avtcore-rtp-over-quic.md
+++ b/draft-ietf-avtcore-rtp-over-quic.md
@@ -80,6 +80,8 @@ blocking).
 
 From time to time, someone asks the reasonable question, "why should anyone implement and deploy RoQ"? This reasonable question deserves a better answer than "because we can". Upon reflection, the following motivations seem useful to state.
 
+The motivations in this section are in no particular order, and this reflects the reality that not all implementers and deployers would agree on "the most important motivations".
+
 ### "Always-On" Transport-level Authentication and Encryption {#alwas-on}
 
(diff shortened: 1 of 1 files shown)


SpencerDawkins:
Just FYI, I went back and thought about the motivations we had for QUIC when we chartered it for HTTP/3, and remembered that some of those motivations also apply for RTP. I also added cross-references to sections later in the document, to make me feel better about not including more details in the motivations themselves. 

Should I add a note that other people likely have other motivations in mind, and it would be great to add them?

PR #120, opened 2023-08-29
Move RTCP tables to appendix and reference section 8

mengelbart:


draft-ietf-avtcore-rtp-over-quic.md
```diff
@@ -1429,6 +1290,150 @@ RoQ, but are not required by RoQ.
...
+# Considered RTCP Packet Types and RTP Header Extensions
+
+This section lists all the RTCP packet types and RTP header extensions that were
+considered in the analysis described in {{rtcp-mapping}}.
+
+## RTCP Control Packet Types {#control-packets}
+
+Several but not all of these control packets and their attributes can be mapped
+from QUIC, as described in {{transport-layer-feedback}}. *Mappable from QUIC*
+has one of three values: *yes*, *QUIC extension required*, and *no*.
```
SpencerDawkins, line 1302: Two things for this text: 

> Several but not all of these control packets and their attributes can be mapped
> from QUIC, as described in {{transport-layer-feedback}}. *Mappable from QUIC*
> has one of three values: *yes*, *QUIC extension required*, and *no*.

I THINK this paragraph applies to more subsections than **## RTCP Control Packet Types {#control-packets}**.

Also, I'm also seeing **partly** and **possibly** in this section, but I'm not seeing **QUIC extension required**.

SpencerDawkins (approved):

draft-ietf-avtcore-rtp-over-quic.md
mengelbart: Thanks! I moved the paragraph to the section introduction so it applies to all the following subsections. I also replaced *QUIC extension required* with *possibly* and *partly*, including short explanations and updated the table values to match their respective descriptions.

PR #123, opened 2023-09-01
Describe considerations for streams, datagrams, and a mixture

### LINKED ISSUE #93: Using streams, datagrams, and per-RTP-frame streams with RoQ
We may need a short section explaining the ROQ-specific tradeoffs

SpencerDawkins:
closes #93 

Note that the "- No link definition for link ID 'rfcnnnn' warnings are fixed in  PR #121


draft-ietf-avtcore-rtp-over-quic.md
```diff
@@ -303,6 +303,23 @@ metrics can be used to generate the required feedback at the sender-side and
...
+* DATAGRAM frames do inherit the QUIC connection's congestion controller. This means that although there is no frame-level flow control, DATAGRAM frames may be delayed until the controller allows them to be sent, or dropped (with an optional notification to the sending application). Implementations can also delay sending DATAGRAM frames to maintain consistent packet pacing (as described in {{Section 7.7 of ?RFC9002}}), and can allow an application to specify a sending expiration time, but these capabilities are not mandated by the standard and may not be present in all implementations.
+* DATAGRAM frames cannot be fragmented. They are limited in size by the max_datagram_frame_size transport parameter, and further limited by the max_udp_payload_size transport parameter and the Maximum Transmission Unit (MTU) of the path between endpoints.
+* DATAGRAM frames belong to a QUIC connection as a whole. There is no QUIC-level way to multiplex/demultiplex DATAGRAM frames within a single QUIC connection. Any multiplexing identifiers must be added, interpreted, and removed by an application, and they will be sent as part of the DATAGRAM frame itself.
```
mengelbart, line 321:
```suggestion
* DATAGRAM frames belong to a QUIC connection as a whole. There is no QUIC-level way to multiplex/demultiplex DATAGRAM frames within a single QUIC connection. Any multiplexing identifiers must be added, interpreted, and removed by an application, and they will be sent as part of the payload of the DATAGRAM frame itself.
```

mengelbart (approved):

draft-ietf-avtcore-rtp-over-quic.md
mengelbart: I think adding that it is part of the payload makes it clear that it will not be in any header field of the frame, like the length would be, for example.

PR #125, opened 2023-09-22
Add multihop topologies considerations for RTCP

mengelbart:


draft-ietf-avtcore-rtp-over-quic.md
```diff
@@ -951,6 +951,22 @@ RTP packet loss can seem lower than actual packet loss due to QUIC's automatic
...
+be able to infer feedback from QUIC, and they may receive less RTCP feedback
+than expected. On the other hand, participants using QUIC may not be aware of
+participants not using QUIC and send as little RTCP as possible since they
+assume their RoQ peer will be able to infer statistics from QUIC. There are two
+ways to solve this problem: Either the translator, translating between RoQ and
+RTP over other protocols such as UDP or TCP, can add RTCP packets for the
+participants not using RoQ by using the statistics it gets from QUIC and the
+mappings described in the following sections. Alternatively, participants may
+use additional signalling to let the RoQ participants know what RTCP is
+required.
```
SpencerDawkins, line 968: This is all correct, but perhaps 

> There are two
> ways to solve this problem: **if the middlebox** translating between RoQ and
> RTP over **other RTP transport** protocols such as UDP or TCP **provides Back-to-Back RTP sessions as described in {{Section 3.2.2 of !RFC7667}}, this middlebox** can add RTCP packets for the
> participants not using RoQ by using the statistics **the middlebox** gets from QUIC and the
> mappings described in the following sections. **If the middlebox does not provide Back-to-Back RTP sessions,** participants may
> use additional signalling to let the RoQ participants know what RTCP is
> required.

would be better?

SpencerDawkins (approved):

PR #132, opened 2023-10-09
Improve BYE to CONNECTION_CLOSE mapping considerations

mengelbart:


draft-ietf-avtcore-rtp-over-quic.md
```diff
@@ -1055,9 +1055,14 @@ frames as described in {{!I-D.draft-smith-quic-receive-ts}} or
...
-QUIC connection. However, while QUIC allows application layer error codes, they
-do not carry string values.
+RTP session participants can use *Goodbye* RTCP packets (`PT=203`, `Name=BYE`,
+{{!RFC3550}}), to indicate that a source is no longer active. If the participant
+is also going to close the QUIC connection, the *BYE* packet can be replaced by
+a QUIC CONNECTION_CLOSE frame. In this case, the reason for leaving can be
+transmitted in QUIC's CONNECTION_CLOSE *Reason Phrase*. However, if the
+participants only wish to leave one RTP session while continuing to use the QUIC
+connection for other data streams, a CONNECTION_CLOSE cannot be used, as it
+would close the QUIC connection.
```
SpencerDawkins, line 1065: I'd suggest rephrasing this sentence as 

> However, if the
> participant only wishes to leave one RTP session while continuing to use the QUIC
> connection, the *BYE* packet MUST be used, because the QUIC CONNECTION_CLOSE 
> would close the entire QUIC connection for all other QUIC streams and datagrams.

SpencerDawkins (approved):

draft-ietf-avtcore-rtp-over-quic.md
mengelbart: I would like to avoid the normative *MUST* in *the BYE packet MUST be used*, because that is not a requirement of RFC 3550 in the first place (*[A participant [...] MAY leave the group without sending a BYE at all.](https://www.rfc-editor.org/rfc/rfc3550#section-6.3.7)*). How about this:

> However, if the
> participant only wishes to leave one RTP session while continuing to use the QUIC
> connection, it has to use the BYE packet because the QUIC CONNECTION_CLOSE
> would close the entire QUIC connection for all other QUIC streams and datagrams.

SpencerDawkins: Hi, @mengelbart - we're getting close to converging. 

The point I was struggling with, is that if you don't send a BYE packet, the sender would continue to send RTP media until the sender detects that the receiver is no longer sending RRs, and "times the receiver out". I agree with you, that inserting a MUST is adding a requirement that isn't in RTP today, so I guess the assumption is that the receiver has some reason for not using BYE, and I shouldn't complain about that. 

I do think it's worth pointing out that we're allowing multiplexing of RTP/RTCP and other protocols in a single QUIC connection, so maybe it's better to also allow for the case where the participant is leaving the RTP session, but is continuing to use the QUIC connection for "something else", by saying something like 

> However, if the
> participant wishes to **use this QUIC
> connection for any other multiplexed traffic**, the participant has to use the BYE packet because the QUIC CONNECTION_CLOSE
> would close the entire QUIC connection for all other QUIC streams and datagrams.

Does that make sense?

Issue #128, opened 2023-09-24
Congestion Control and Rate Control


aboba:
From the AVTCORE WG mailing list: https://mailarchive.ietf.org/arch/msg/avt/7IFLBJAmFbB3f1_OgkoPyN0lhT8/

Looking over the coverage Congestion Control and Rate Control, the two
topics appear to be conflated and also there appear to be some issues that
have not been fully considered.

Section 1.2.2

While the effect of QUIC's response to congestion means that some RTP
packets will arrive at the receiver later than a user of the RTP flow might
prefer, it is still preferable to "ceasing transmission" completely until
the RTP sender has a reason to believe that restarting the flow will not
result in congestion.¶
<https://datatracker.ietf.org/doc/html/draft-ietf-avtcore-rtp-over-quic-05#section-1.2.2-3>

[BA] In contrast to circuit breakers, which do not restrict the ability to
send RTCP feedback, QUIC congestion control affects RTCP feedback, not just
RTP.  So saying QUIC congestion control is "preferable" seems questionable.

Moreover, when a single QUIC connection is used to multiplex both RTP-RTCP
and non-RTP packets as described in Section 1.2.5
<https://datatracker.ietf.org/doc/html/draft-ietf-avtcore-rtp-over-quic-05#single-path>,
the QUIC connection will still be Internet-safe, with no coordination
required.

[BA] While it may be "Internet-safe", delays in RTCP feedback are likely to
destabilize rate control as well as resulting in challenges to A/V sync.
So not sure that "Internet-safe" is the only important metric here.

Section 1.2.3

One word of caution is in order - RTP implementations may rely on at least
some minimal periodic RTCP feedback, in order to determine that an RTP flow
is still active, and is not causing sustained congestion (as described in [
RFC8083 <https://www.rfc-editor.org/rfc/rfc8083>], but since this
"periodicity" is measured in seconds, the impact of this "duplicate"
feedback on path bandwidth utilization is likely close to zero.

[BA] Under congestion, RTCP feedback can potentially be delayed
substantially. Here is the issue is not "bandwidth utilization" but whether
RTCP receives the transport treatment required for control traffic.  Note
also that similar considerations apply to treatment of audio vs. video.
Serious problems with a/v sync are possible (or even likely) under
congestion.

Section 1.2.4

This is especially useful in certain conferencing topologies, where
otherwise senders have no choice but to use the lowest path MTU for all
conference participants, but even in point-to-point RTP sessions, this also
[...]

PR #133, opened 2023-10-10
Move RTCP mappings requiring extensions to appendix

### LINKED ISSUE #127: Normative References
From the AVTCORE WG mailing list: https://mailarchive.ietf.org/arch/msg/avt/WH8OTXO7LT8IQzxPh2QwfeGVcF0/

With the QUIC specifications still evolving, this specification is put in a difficult position.  Beyond the basic QUIC specifications, there are extensions which could be of potential benefit.  However, some of the extensions are not yet WG work items, or are not widely deployed, or may take some time to be published as RFCs, or may never be published as RFCs.

Currently, the document has an Appendix A: List of Optional QUIC extensions, which states "The following is a list of QUIC protocol extensions that might be beneficial for RoQ, but are not required by RoQ." Putting discussion of non-RFC QUIC extensions into a (non-normative) Appendix seems like a good idea, if it could allow the main specification to focus on the core QUIC functionality and avoid normative references to documents which may or may not ever be published as RFCs.

However, the problem is that discussion of the extensions is not limited to the (non-normative) Appendix, but is contained in other sections, including Sections 7, 7.3.4 and 8.1.

This results in normative references to [I-D.draft-huitema-quic-ts], [I-D.draft-ietf-quic-ack-frequency] and [I-D.draft-smith-quic-receive-ts]. ¶

I would like to suggest that the RTP over QUIC specification be restructured so as to only normatively reference documents which have been published as RFCs.

For this to work, material in Sections 6, 7 and 8 may need to be moved to Appendix A, which would presumably avoid normative references.

mengelbart:
close #127 

This PR does not yet remove the references to the timestamp extensions from section 7. (*Congestion Control and Rate Adaptation*). I am planning to do that in a PR for #128.


joerg-ott (approved):

SpencerDawkins (approved):
This looked fine to me, with no changes requested.

PR #134, opened 2023-10-10
Clarify congestion and rate control

### LINKED ISSUE #128: Congestion Control and Rate Control

mengelbart:
close #128

### DIFF
--- a/draft-ietf-avtcore-rtp-over-quic.md
+++ b/draft-ietf-avtcore-rtp-over-quic.md
@@ -715,7 +715,7 @@ This document discusses aspects of transport level congestion control in
 congestion control algorithm for QUIC or rate adaptation algorithm for RTP.
 
 This document also gives guidance about avoiding problems with *nested*
-congestion controllers in {{nested-CC}}.
+congestion controllers in {{rate-adaptation-application-layer}}.
 
 This document also discusses congestion control implications of using shared or
 multiple separate QUIC connections to send and receive multiple independent data
@@ -799,41 +799,6 @@ before passing data to the QUIC layer.
 The bandwidth estimation algorithm typically needs some feedback on the
 transmission performance. This feedback can be collected via RTCP or following
 the guidelines in {{rtcp-mapping}} and {{api-considerations}}.
-
-## Resolving Interactions Between QUIC and Application-layer Congestion Control {#nested-CC}
-
-Because QUIC is a congestion-controlled transport, as described in
-{{cc-quic-layer}}, and RTP applications can also perform congestion control and
-rate adaptation, as described in {{rate-adaptation-application-layer}},
-implementers should be aware of the possibility that these "nested" congestion
-control loops, where both controllers are managing rate adaptation for the same
-packet stream independently, may deliver problematic performance. Because this
-document is describing a specific case (media transport), we can provide some
-guidance to avoid the worst possible problems.
-
-- **Application-limited Media Flows** - if an application chooses RTP as its
-  transport mechanism, the goal will be maximizing the user experience, not
-  maximizing path bandwidth utilization. If the application is, in fact,
-  transmitting media that does not saturate path bandwidth, and paces its
-  transmission, more heavy-handed congestion control mechanisms (drastic
-  reductions in the sending rate when loss is detected, with much slower
-  increases when losses are no longer detected) should rarely come into play. If
-  the application chooses ROQ as its transport, sends enough media to saturate
-  the path bandwidth, and does not adapt its own sending rate, drastic measures
-  will be required in order to avoid sustained or oscillating congestion along
-  the path.
-
-- **Awareness of Bufferbloat** - modern general-purpose congestion controllers
-  do not adjust  their sending rates based only on packet loss. For example, BBR
-  ("Bottleneck Bandwidth and Round-trip propagation time")
-  {{?I-D.cardwell-iccrg-bbr-congestion-control}} describes its strategy for
-  adapting sending rates in this way:
-
-> (BBR) "uses recent measurements of a transport connection's delivery rate,
-> round-trip time, and packet loss rate to build an explicit model of the
-> network path. BBR then uses this model to control both how fast it sends data
-> and the maximum volume of data it allows in flight in the network at any
-> time."
 
 ## Sharing QUIC connections {#shared-connections}
 
--- a/draft-ietf-avtcore-rtp-over-quic.md
+++ b/draft-ietf-avtcore-rtp-over-quic.md
@@ -712,7 +712,7 @@ environment.
 This document discusses aspects of transport level congestion control in
 {{cc-quic-layer}} and application layer rate control in
 {{rate-adaptation-application-layer}}. It does not mandate any specific
-congestion control or rate adaptation algorithm for QUIC or RTP.
+congestion control algorithm for QUIC or rate adaptation algorithm for RTP.
 
 This document also gives guidance about avoiding problems with *nested*
 congestion controllers in {{nested-CC}}.
--- a/draft-ietf-avtcore-rtp-over-quic.md
+++ b/draft-ietf-avtcore-rtp-over-quic.md
@@ -252,12 +252,12 @@ QUIC congestion controller:
 Rate Adaptation:
 : A mechanism that adjusts the sending rate of an application in order to
 maximize the amount of information that is sent to a receiver without causing
-queues to build beyond a reasonable amount, causing *buffer bloat* and *jitter*.
-Rate adaptation is one way to respond to send rate limitations imposed by
-congestion control algorithms. When a sender has multiple media streams to the
-receiver, the sum of all sending rates for media streams must not be high enough
-to cause congestion on the path these media streams share between sender and
-receiver.
+buffer bloat, when queues build beyond a reasonable amount, or jitter, when
+interpacket arrival times fluctuate due to queuing delays. Rate adaptation is
+one way to respond to sending rate limitations imposed by congestion control
+algorithms. When a sender has multiple media streams to the receiver, the sum of
+all sending rates for media streams must not be high enough to cause congestion
+on the path these media streams share between sender and receiver.
 
 Receiver:
 : An endpoint that receives media in RTP packets and may send or receive RTCP packets.
@@ -723,9 +723,9 @@ congestion control or rate adaptation algorithm for QUIC or RTP.
 This document also gives guidance about avoiding problems with *nested*
 congestion controllers in {{nested-CC}}.
 
-{{shared-connections}} discusses congestion control implications of using shared
-or multiple separate QUIC connections to send and receive multiple independent
-data streams.
+This document also discusses congestion control implications of using shared or
+multiple separate QUIC connections to send and receive multiple independent data
+streams in {{shared-connections}}.
 
 ## Congestion Control at the Transport Layer {#cc-quic-layer}
 
@@ -766,10 +766,11 @@ implement an alternative bandwidth estimation at the application layer as
 described in {{rate-adaptation-application-layer}}.
 
 It is assumed that the congestion controller in use provides a pacing mechanism
-to determine when a packet can be sent to avoid bursts. The currently proposed
-congestion control algorithms for real-time communications (e.g., SCReAM and
-NADA) provide such pacing mechanisms, and QUIC recommends pacing for senders
-based on the congestion control algorithm.
+to determine when a packet can be sent to avoid bursts and minimize variation in
+inter-packet arrival times. The currently proposed congestion control algorithms
+for real-time communications (e.g., SCReAM and NADA) provide such pacing
+mechanisms, and QUIC recommends pacing for senders based on the congestion
+control algorithm.
 
 ## Rate Adaptation at the Application Layer {#rate-adaptation-application-layer}
 
@@ -792,7 +793,7 @@ controllers.
 If the application transmits media that does not saturate path bandwidth and
 paces its transmission, more heavy-handed congestion control mechanisms (drastic
 reductions in the sending rate when loss is detected, with much slower increases
-when losses are no longer detected) should rarely come into play. If the
+when losses are no longer being detected) should rarely come into play. If the
 application chooses RoQ as its transport, sends enough media to saturate the
 path bandwidth, and does not adapt it's sending rate, drastic measures will be
 required to avoid sustained or oscillating congestion along the path.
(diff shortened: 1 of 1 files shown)


draft-ietf-avtcore-rtp-over-quic.md
```diff
@@ -711,111 +716,112 @@ CONNECTION\_CLOSE frame with one of the error codes defined in
...
+The application needs a mechanism to query the available bandwidth to adapt
+media codec configurations. If the employed congestion controller of the QUIC
+connection keeps an estimate of the available bandwidth, it could expose an API
+to the application to query the current estimate. If the congestion controller
+cannot provide a current bandwidth estimate to the application, the sender can
+implement an alternative bandwidth estimation at the application layer as
+described in {{rate-adaptation-application-layer}}.
+
+It is assumed that the congestion controller in use provides a pacing mechanism
+to determine when a packet can be sent to avoid bursts. The currently proposed
```
SpencerDawkins, line 776: I'd suggest 

> to avoid bursts and minimize variation in inter-packet arrival times.

draft-ietf-avtcore-rtp-over-quic.md
```diff
@@ -711,111 +716,112 @@ CONNECTION\_CLOSE frame with one of the error codes defined in
...
+the rate of the observed available bandwidth. Applications can use the same
+bandwidth estimation to adapt their rate when using QUIC. However, running an
+additional congestion control algorithm at the application layer can have
+unintended effects due to the interaction of two *nested* congestion
+controllers.
+
+If the application transmits media that does not saturate path bandwidth and
+paces its transmission, more heavy-handed congestion control mechanisms (drastic
+reductions in the sending rate when loss is detected, with much slower increases
+when losses are no longer detected) should rarely come into play. If the
```
SpencerDawkins, line 802: I'd suggest

> when losses are no longer **being** detected)

SpencerDawkins (approved):

SpencerDawkins (approved):
This is ready to merge.

PR #138, opened 2023-10-11
Remove normative API requirements

mengelbart:
I don't think normative SHOULD's are useful for describing API considerations. This also clarifies the datagram/stream acknowledgement mappings.

Additionally, I removed the API write functions section. I think resetting streams is a core feature of QUIC and it shouldn't be necessary to be listed explicitly. And congestion control configuration is not something that needs an API at runtime. The updates to the congestion control section in #134 should also make this more clear.


SpencerDawkins (approved):
This looks fine.

PR #139, opened 2023-10-11
Allow to resume sending datagrams after STOP_SENDING

mengelbart:


joerg-ott (changes requested):
or as datagram(s)?

draft-ietf-avtcore-rtp-over-quic.md
```diff
@@ -587,9 +587,10 @@ frame MUST be set to ROQ\_FRAME\_CANCELLED.
 
 When a RoQ sender receives a STOP\_SENDING frame for the last open stream
 available to send RTP/RTCP-data, the RoQ sender MUST open one or more new QUIC
-streams before sending new media frames. Any media frame that has already been
-sent on the QUIC stream that received the STOP\_SENDING frame, MUST NOT be sent
-again on the new QUIC stream(s).
+streams before sending new media frames or continue to send new media frames in
+QUIC datagrams. Any media frame that has already been sent on the QUIC stream
+that received the STOP\_SENDING frame, MUST NOT be sent again on the new QUIC
+stream(s) or datagrams.
```
SpencerDawkins, line 593: Mentioning datagrams within the section discussing STOP\_SENDING seems confusing, because this is in Section 5.2 (streams), and because datagrams aren't associated with a stream, STOP\_SENDING and RESET\_STREAM won't have any effect on whether the sender continues to send datagrams. 

I'm not sure what you're trying to get at, with this change. I might guess, but I'd be guessing. HELP! :no_mouth:

(Ping me on Slack if we need to talk, please!)

SpencerDawkins (commented):
Mentioning datagrams within the section discussing STOP_SENDING seems confusing, because this is in Section 5.2 (streams), and because datagrams aren't associated with a stream, STOP_SENDING and RESET_STREAM won't have any effect on whether the sender continues to send datagrams.

If you want to say the same kind of thing about datagrams that we're saying here about streams ("don't re-send old media"), I guess we can say that in Section 5.3, but I'm not sure I understand why we need to say anything. 

But I'm not sure what you're trying to get at, with this change. I might guess, but I'd be guessing. HELP! 😶

(Ping me on Slack if we need to talk, please!)

mengelbart:
The problem I see is with this sentence:

> the RoQ sender MUST open one or more new QUIC streams before sending new media frames.

To me, this sounds like I definitely MUST continue sending new media on new streams. But I might want to continue sending the subsequent frames as datagrams instead (and never open a new QUIC stream again). We might instead add something like *STOP_SENDING is not a request to end the media stream but only an indication that a receiver stopped reading a media frame. A sender SHOULD continue sending new media frames of the same media stream. New media frames can be sent in new QUIC streams or as QUIC datagrams (see section ...).*

I'm using *SHOULD* here because the sender may decide to stop the media stream for different reasons.

SpencerDawkins:
@mengelbart - we're converging. I have a somewhat different suggestion, but I agree with where you're going with this text. 

> STOP_SENDING is not a request **to the sender to stop sending the RTP media stream, only an indication that a receiver stopped reading the QUIC stream being used**. A sender **with additional media frames to send SHOULD continue sending them on a QUIC stream in the "Ready" state**. New media frames can be sent in new QUIC streams or as QUIC datagrams (see section ...).

Does that make sense?

mengelbart:
> > STOP_SENDING is not a request **to the sender to stop sending the RTP media stream, only an indication that a receiver stopped reading the QUIC stream being used**. A sender **with additional media frames to send SHOULD continue sending them on a QUIC stream in the "Ready" state**. **Alternatively**, new media frames can be sent as QUIC datagrams (see section ...).


I updated the PR, but instead of `a QUIC stream in the "Ready" state`, I used `another QUIC stream,` which is closer to how it was before the change because the *Ready* state is only one of the possible states described in RFC9000. I think it could also be in the *Send* state if the other QUIC stream was used for other packets before.

SpencerDawkins:
Hi, @mengelbart - I think we're converged here (on your text). I was hoping I would be able to qualify "another QUIC stream", but after reflection, I don't think that's necessary or helpful. 

Dude - merge it!

PR #140, opened 2023-10-11
Add IANA considerations for error codes

### LINKED ISSUE #76: Important: define error codes
Related to #13 but also applicable to the rest of the specification.

RTP over QUIC is an application layer mapping for RTP applications. 

When using STOP_SENDING or RESET_STREAM, the code has to be an application-layer error code, and you don't have one. Without that, people will just pick random numbers (or worse, send 0 for everything). 

Defining codes isn't too hard, we should put some thought into gradeful or abrupt conditions that can be communicated via error codes

mengelbart:
closes #76

### DIFF
--- a/draft-ietf-avtcore-rtp-over-quic.md
+++ b/draft-ietf-avtcore-rtp-over-quic.md
@@ -1289,7 +1289,7 @@ Approval as defined in {{Sections 4.9 and 4.10 of !RFC8126}}.
 
 Registrations for error codes are required to include a description of the error
 code. An expert reviewer is advised to examine new registrations for possible
-duplication with existing error codes.
+duplication or interaction with existing error codes.
 
 In addition to common fields as described in Section {{Section 22.1 of
 !RFC9000}}, this registry includes two additional fields. Permanent
@@ -1299,7 +1299,8 @@ Name:
 : A name for the error code.
 
 Description:
-: A brief description of the error code semantics.
+: A brief description of the error code semantics, which MAY be a summary if a
+specification reference is provided.
 
 The initial allocations in this registry are all assigned permanent status and
 list a change controller of the IETF and a contact of the AVTCORE working group
--- a/draft-ietf-avtcore-rtp-over-quic.md
+++ b/draft-ietf-avtcore-rtp-over-quic.md
@@ -1307,16 +1307,16 @@ list a change controller of the IETF and a contact of the AVTCORE working group
 
 The entries in {{tab-error-codes}} are registered by this document.
 
-| Name                          | Value  | Description                            | Specification      |
-| ----------------------------- | ------ | -------------------------------------- | ------------------ |
-| ROQ\_NO\_ERROR                | 0x01 | No Error                               | {{error-handling}} |
-| ROQ\_GENERAL\_ERROR           | 0x02 | General error                          | {{error-handling}} |
-| ROQ\_INTERNAL\_ERROR          | 0x03 | Internal Error                         | {{error-handling}} |
-| ROQ\_PACKET\_ERROR            | 0x04 | Invalid payload format                 | {{error-handling}} |
-| ROQ\_STREAM\_CREATION\_ERROR  | 0x05 | Invalid stream type                    | {{error-handling}} |
-| ROQ\_FRAME\_CANCELLED         | 0x06 | Frame cancelled                        | {{error-handling}} |
-| ROQ\_UNKNOWN\_FLOW\_ID        | 0x07 | Unknown Flow ID                        | {{error-handling}} |
-| ROQ\_EXPECTATION\_UNMET       | 0x08 | Externally signalled requirement unmet | {{error-handling}} |
+| Value  | Name                          | Description                            | Specification      |
+| ------ | ----------------------------- | -------------------------------------- | ------------------ |
+| 0x01   | ROQ\_NO\_ERROR                | No Error                               | {{error-handling}} |
+| 0x02   | ROQ\_GENERAL\_ERROR           | General error                          | {{error-handling}} |
+| 0x03   | ROQ\_INTERNAL\_ERROR          | Internal Error                         | {{error-handling}} |
+| 0x04   | ROQ\_PACKET\_ERROR            | Invalid payload format                 | {{error-handling}} |
+| 0x05   | ROQ\_STREAM\_CREATION\_ERROR  | Invalid stream type                    | {{error-handling}} |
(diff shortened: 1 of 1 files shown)


joerg-ott (approved):

draft-ietf-avtcore-rtp-over-quic.md
```diff
@@ -1270,20 +1274,51 @@ The "rtp-mux-quic" string identifies RoQ:
...
   : This document
 
-## RoQ Error Codes
+## RoQ Error Codes Registry {#iana-error-codes}
+
+This document establishes a registry for RoQ error codes. The "RoQ Error Codes"
+registry manages a 62-bit space.
+
+The new error codes registry created in this document operates under the QUIC
+registration policy documented in {{Section 22.1 of !RFC9000}}. This registry
```
SpencerDawkins, line 1283: I don't understand this part at all. There's a lot of cool stuff in 22.1, but is it all applicable to the RoQ error code registry?

draft-ietf-avtcore-rtp-over-quic.md
```diff
@@ -1270,20 +1274,51 @@ The "rtp-mux-quic" string identifies RoQ:
...
+includes the common set of fields listed in {{Section 22.1.1 of !RFC9000}}.
+
+Permanent registrations in this registry are assigned using the Specification
+Required policy ({{!RFC8126}}), except for values between 0x00 and 0x3f (in
+hexadecimal; inclusive), which are assigned using Standards Action or IESG
+Approval as defined in {{Sections 4.9 and 4.10 of !RFC8126}}.
+
+Registrations for error codes are required to include a description of the error
+code. An expert reviewer is advised to examine new registrations for possible
+duplication with existing error codes.
```
SpencerDawkins, line 1293: Possibly 

> duplication **or interaction** with existing error codes

draft-ietf-avtcore-rtp-over-quic.md
```diff
@@ -1270,20 +1274,51 @@ The "rtp-mux-quic" string identifies RoQ:
...
+
+Description:
+: A brief description of the error code semantics.
+
+The initial allocations in this registry are all assigned permanent status and
+list a change controller of the IETF and a contact of the AVTCORE working group
+(avt@ietf.org).
 
-This document establishes a registry for RoQ error codes.
+The entries in {{tab-error-codes}} are registered by this document.
```
SpencerDawkins, line 1309: Especially because we're tying this registry to QUIC practice, the initial registration should probably contain all the fields for each entry, in the order we want them to appear, as the QUIC registries [here](https://www.iana.org/assignments/quic/quic.xhtml#quic-transport-error-codes) show.

draft-ietf-avtcore-rtp-over-quic.md
```diff
@@ -1270,20 +1274,51 @@ The "rtp-mux-quic" string identifies RoQ:
...
+
+Registrations for error codes are required to include a description of the error
+code. An expert reviewer is advised to examine new registrations for possible
+duplication with existing error codes.
+
+In addition to common fields as described in Section {{Section 22.1 of
+!RFC9000}}, this registry includes two additional fields. Permanent
+registrations in this registry MUST include the following fields:
+
+Name:
```
SpencerDawkins, line 1299: If you want to match the [QUIC transport error code registry](https://www.iana.org/assignments/quic/quic.xhtml#quic-transport-error-codes), this should be "CODE".

draft-ietf-avtcore-rtp-over-quic.md
```diff
@@ -1270,20 +1274,51 @@ The "rtp-mux-quic" string identifies RoQ:
...
+: A name for the error code.
```
SpencerDawkins, line 1300: > A short mnemonic for the parameter

draft-ietf-avtcore-rtp-over-quic.md
```diff
@@ -1270,20 +1274,51 @@ The "rtp-mux-quic" string identifies RoQ:
...
+
+Description:
+: A brief description of the error code semantics.
```
SpencerDawkins, line 1303: We might as well match this, too:

> A brief description of the error code semantics, which MAY be a summary if a specification reference is provided.

SpencerDawkins (approved):

draft-ietf-avtcore-rtp-over-quic.md
mengelbart: I tried to follow what [HTTP/3](https://datatracker.ietf.org/doc/html/rfc9114#name-error-codes) and [DNS over QUIC](https://datatracker.ietf.org/doc/html/rfc9250#name-dns-over-quic-error-codes-r) did. I'm not an expert on this and it may not be applicable, but to me, it seemed to make sense to follow what other application protocol mappings to QUIC do.

draft-ietf-avtcore-rtp-over-quic.md
mengelbart: I don't mind what we use here, but HTTP/3 uses *Name*, and DNS over QUIC uses *Error* :D

draft-ietf-avtcore-rtp-over-quic.md
mengelbart: Also taken from HTTP/3. I find *A name for the error code* much easier to understand, especially since the codes are not called parameters anywhere else.

draft-ietf-avtcore-rtp-over-quic.md
mengelbart: I think all fields are included: *Name*, *Value*, *Description*, and *Specification* are in the table, *Status*, *Change Controller*, and *Contact* for all are given in the text above. HTTP/3 uses this order but the table in the registry uses a different order. I'll swap it here, so we may have the same order here and in the registry.

SpencerDawkins:
@mengelbart, this seems ready to merge - thank you for looking over my comments!

draft-ietf-avtcore-rtp-over-quic.md
SpencerDawkins: As we discussed yesterday, I knew we were trying to match the QUIC error codes, but didn't realize that HTTP/3 error codes and DNS over QUIC error codes didn't match that order, or each other. My apologies, obviously my comment doesn't matter!

SpencerDawkins (commented):
I see a couple of nits that would be good to fix before we call everyone's attention to the IANA section. Could you let me know in Slack when you've had a chance to fix them?

Issue #111, opened 2023-07-23
Check SHOULD requirements for ROQ receivers 


SpencerDawkins:
This is probably premature, but I noticed 

> To send RTP/RTCP packets over QUIC streams, a sender MUST open a
> new unidirectional QUIC stream. Streams are unidirectional because there is no
> synchronous relationship between sent and received RTP/RTCP packets. A peer that
> receives a bidirectional stream with a flow identifier that is associated with
> an RTP or RTCP stream, **SHOULD** stop reading from the stream and send a
> STOP\_SENDING frame with the application protocol error code set to
> ROQ\_STREAM\_CREATION\_ERROR.

As I noted in comments on PR #91, that's a fundamental violation of the protocol ("MUST open a new unidirectional QUIC stream"). If that's a MUST, ISTM that we need to either make detecting bidirectional ROQ connections a MUST, or explain what would be likely to happen if the receiver just decides to read from the stream and - what - never write to it? or start writing to it, in spite of the stream violating the protocol? 

But this is actually just a reminder that we need to look at SHOULDs for receivers, and make sure that it makes sense for a receiver to ignore the SHOULD action, and explain at least some of the likely consequences if the receiver does so. 

As long as we're still writing significant normative text, it may make sense to wait and check all the SHOULDs at once, later in the process. As long as we don't forget.

SpencerDawkins:
I think we've written enough text to justify looking at BCP 14 language, and especially at SHOULDs. I'll start on a PR for that.

PR #143, opened 2023-10-18
Configure streams

### LINKED ISSUE #142: Add text for configuring unidirectional RoQ streams
A LONG time ago, in [an issue far, far away](https://github.com/mengelbart/rtp-over-quic-draft/issues/13#issuecomment-1172980290), @LPardue said

You're right about zero credits, that could always happen. HTTP/3 recommends at least 100 concurrent bidirectional streams at any time https://www.rfc-editor.org/rfc/rfc9114.html#section-6.1-2. Section 6.1 and 6.2 more broadly discuss considerations for flow control and unidirectional streams. I'd suggest other application mappings think of similar guidance.

That's gotten a bit lost, up until now, and it doesn't have anything to do with datagrams. I'm moving that to its own issue (here)

SpencerDawkins:
closes #142

### DIFF
--- a/draft-ietf-avtcore-rtp-over-quic.md
+++ b/draft-ietf-avtcore-rtp-over-quic.md
@@ -507,10 +507,6 @@ of flows.
 ## QUIC Streams {#quic-streams}
 
 To send RTP/RTCP packets over QUIC streams, a sender MUST open at least one new unidirectional QUIC stream.
-In order to permit QUIC streams to open, a RoQ sender SHOULD configure non-zero minimum values for the number of permitted streams and the initial stream flow-control window, based on the number of parallel, or simultaneously active, RTP/RTCP flows.
-Endpoints that excessively restrict the number of streams or the flow-control window of these streams will increase the chance that the remote peer reaches the limit early and becomes blocked.
-More discussion about selecting these values appears in {{quic-flow-cc}}.
-
 RoQ uses unidirectional streams, because there is no
 synchronous relationship between sent and received RTP/RTCP packets. A peer that
 receives a bidirectional stream with a flow identifier that is associated with
@@ -623,6 +619,9 @@ retransmitted by QUIC.
 
 ### Flow control and MAX\_STREAMS {#quic-flow-cc}
 
+In order to permit QUIC streams to open, a RoQ sender SHOULD configure non-zero minimum values for the number of permitted streams and the initial stream flow-control window, based on the number of parallel, or simultaneously active, RTP/RTCP flows.
+Endpoints that excessively restrict the number of streams or the flow-control window of these streams will increase the chance that the remote peer reaches the limit early and becomes blocked.
+
 Opening new streams for new packets MAY implicitly limit the number of packets
 concurrently in transit because the QUIC receiver provides an upper bound of
 parallel streams, which it can update using QUIC MAX\_STREAMS frames. The number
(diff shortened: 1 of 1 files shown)


draft-ietf-avtcore-rtp-over-quic.md
```diff
@@ -506,8 +506,12 @@ of flows.
 
 ## QUIC Streams {#quic-streams}
 
-To send RTP/RTCP packets over QUIC streams, a sender MUST open a
-new unidirectional QUIC stream. Streams are unidirectional because there is no
+To send RTP/RTCP packets over QUIC streams, a sender MUST open at least one new unidirectional QUIC stream.
+In order to permit QUIC streams to open, a RoQ sender SHOULD configure non-zero minimum values for the number of permitted streams and the initial stream flow-control window, based on the number of parallel, or simultaneously active, RTP/RTCP flows.
```
mengelbart, line 510: This is a useful addition, but I suggest moving it into the MAX_STREAMS/flow control subsection to keep all the things related to that issue in one place.

SpencerDawkins: Agreed (thanks for the helpful suggestion). I committed a change for this.

SpencerDawkins:
@mengelbart - I'm looking at this text, also in {#quic-streams},

> A RoQ receiver MUST be prepared to receive RTP packets on any number of QUIC streams (subject to its limit on parallel open streams) and SHOULD not make assumptions about which RTP sequence numbers are carried in which streams.
> 
> Note: A sender may or may not decide to discontinue using a lower stream number after starting packet transmission on a higher stream number.

I think this would be better as 

> **Because a sender can continue sending on a lower stream number after starting packet transmission on a higher stream number,** a RoQ receiver MUST be prepared to receive RoQ packets on any number of QUIC streams (subject to its limit on parallel open streams) and SHOULD not make assumptions about which RTP sequence numbers are carried in which streams.

I'm not wild about  

> **SHOULD** not make assumptions about which RTP sequence numbers are carried in which streams

for a couple of reasons.

Is this more like 

> SHOULD be prepared to accept any RTP sequence number on any open stream?

We can wait to talk about the SHOULD until we work on #111, but in any event, 

- we should have a legitimate reason for why a RoQ receiver would choose not to accept any RTP sequence number on any open stream, given that a RoQ sender can send any RTP sequence number on any open stream, and 
- if a RoQ receiver makes that choice, we need to think about what happens next.

If you send me an RTP packet with sequence number 12 on stream 32, and I'm not prepared to accept it. Do I drop it? Do I need to tell you that I'm dropping it because of the stream it arrived on, so you don't look at the ACKs/RRs and decide that you need to reduce the sending rate (because you're sending RTP packets that aren't being acknowledged)? Etc.

mengelbart:
> @mengelbart - I'm looking at this text, also in {#quic-streams},
> 
> > A RoQ receiver MUST be prepared to receive RTP packets on any number of QUIC streams (subject to its limit on parallel open streams) and SHOULD not make assumptions about which RTP sequence numbers are carried in which streams.
> > Note: A sender may or may not decide to discontinue using a lower stream number after starting packet transmission on a higher stream number.
> 
> I think this would be better as
> 
> > **Because a sender can continue sending on a lower stream number after starting packet transmission on a higher stream number,** a RoQ receiver MUST be prepared to receive RoQ packets on any number of QUIC streams (subject to its limit on parallel open streams) and SHOULD not make assumptions about which RTP sequence numbers are carried in which streams.

Sounds good to me.

> I'm not wild about
> 
> > **SHOULD** not make assumptions about which RTP sequence numbers are carried in which streams
> 
> for a couple of reasons.
> 
> Is this more like
> 
> > SHOULD be prepared to accept any RTP sequence number on any open stream?
> 
> We can wait to talk about the SHOULD until we work on #111, but in any event,
> 
>     * we should have a legitimate reason for why a RoQ receiver would choose not to accept any RTP sequence number on any open stream, given that a RoQ sender can send any RTP sequence number on any open stream, and
> 
>     * if a RoQ receiver makes that choice, we need to think about what happens next.
> 
> 
> If you send me an RTP packet with sequence number 12 on stream 32, and I'm not prepared to accept it. Do I drop it? Do I need to tell you that I'm dropping it because of the stream it arrived on, so you don't look at the ACKs/RRs and decide that you need to reduce the sending rate (because you're sending RTP packets that aren't being acknowledged)? Etc.

These are good points. I think we want to use *MUST* instead. There should be no special treatment of RTP packets just because they arrive on a stream with a lower ID or on datagrams.

PR #144, opened 2023-10-19
Describe strategies for reacting to datagram non-support

### LINKED ISSUE #13: State what endpoints should do if the QUIC datagram extension is not enabled
The RTP over QUIC puts a hard requirement on endpoints advertising their support for the QUIC datagram extension.

If endpoints use ALPN to select RTP over QUIC, but the QUIC transport parameters do not contain a `max_datagram_frame_size`, it is not clear what should happen. One approach would be to fail the connection at the point this is detected, such as rejcting the connection handshake. 

Not stating an expected way to handle this could lead to some interop problems.

### LINKED ISSUE #122: Define how RoQ reacts to the absence of max_datagram_frame_size transport parameter
While working on a PR for Issue #93 I noticed the following text in [Section 3 of RFC 9221](https://datatracker.ietf.org/doc/html/rfc9221#name-transport-parameter):

> Application protocols that use datagrams **MUST** define how they react to the absence of the max_datagram_frame_size transport parameter. If datagram support is integral to the application, the application protocol can fail the handshake if the max_datagram_frame_size transport parameter is not present.

I don't think this belongs in a subsection of [Section 3 in the RoQ draft](https://mengelbart.github.io/rtp-over-quic-draft/draft-ietf-avtcore-rtp-over-quic.html#name-protocol-overview), but we do need to do it, and we should talk about what the behavior should be before we add this to the draft (probably in [Section 4 of the ROQ draft](https://mengelbart.github.io/rtp-over-quic-draft/draft-ietf-avtcore-rtp-over-quic.html#name-connection-establishment-an).

SpencerDawkins:
Please note that I moved text about QUIC implementations exposing maximum datagram sizes to the API section as well. 

closes #122 
closes #13

### DIFF
--- a/draft-ietf-avtcore-rtp-over-quic.md
+++ b/draft-ietf-avtcore-rtp-over-quic.md
@@ -272,10 +272,7 @@ This document introduces a mapping of the Real-time Transport Protocol (RTP) to
 the QUIC transport protocol. RoQ allows the use of QUIC streams and
 QUIC datagrams to transport real-time data, and thus, the QUIC
 implementation MUST support QUIC's datagram extension, if RTP packets
-should be sent over QUIC datagrams. Since datagram frames cannot be fragmented,
-the QUIC implementation MUST also provide a way to query the maximum datagram
-size so that an application can create RTP packets that always fit into a QUIC
-datagram frame.
+should be sent over QUIC datagrams.
 
 {{!RFC3550}} specifies that RTP sessions need to be transmitted on different
 transport addresses to allow multiplexing between them. RoQ uses a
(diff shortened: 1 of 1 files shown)


draft-ietf-avtcore-rtp-over-quic.md
```diff
@@ -1152,8 +1154,7 @@ This section provides a list of items that an application might want to export
 from an underlying QUIC implementation. It is thus RECOMMENDED that a QUIC
 implementation exports the listed items to the application.
 
-* *Maximum Datagram Size*: The maximum datagram size that the QUIC connection
-  can transmit.
+* *Maximum Datagram Size*: The maximum datagram size that the QUIC connection can transmit. Since datagram frames cannot be fragmented, the QUIC implementation MUST also provide a way to query the maximum datagram size so that an application can create RTP packets that always fit into a QUIC datagram frame.
```
mengelbart, line 1157: I would prefer to keep the normative MUST out of here. #138 removes all the other normative SHOULDs from this section, too. I would suggest removing the MUST here and in the datagram encapsulation section. Instead, we should explain why the application should be able to learn about the maximum size and what can happen if it cannot. If the application does not know the maximum size, it may try to send datagrams that will be dropped. I think it makes sense to explain that in the datagram encapsulation section.

mengelbart (changes requested):

draft-ietf-avtcore-rtp-over-quic.md
SpencerDawkins: > I would prefer to keep the normative MUST out of here. #138 removes all the other normative SHOULDs from this section, too. I would suggest removing the MUST here and in the datagram encapsulation section. Instead, we should explain why the application should be able to learn about the maximum size and what can happen if it cannot. If the application does not know the maximum size, it may try to send datagrams that will be dropped. I think it makes sense to explain that in the datagram encapsulation section.

@mengelbart - that makes sense to me, and I'll make the change you're suggesting, but I should point out that 

> This section provides a list of items that an application might want to export
from an underlying QUIC implementation. It is thus **RECOMMENDED** that a QUIC
implementation exports the listed items to the application.

is also BCP14-normative ... our terminology section uses the BCP14 template, and [BCP14 uses SHOULD and RECOMMENDED as synonyms](https://datatracker.ietf.org/doc/html/rfc2119#autoid-3). 

Would you like me to make a change there, too?

mengelbart: No need to. It will already be changed with #138.

PR #146, opened 2023-10-20
Improve congestion control  and rate adaptation in motivation section

### LINKED ISSUE #135: Congestion and rate control in motivation section
From @aboba in https://github.com/mengelbart/rtp-over-quic-draft/issues/128:

> 
> Looking over the coverage of Congestion Control and Rate Control, the two
> topics appear to be conflated in places and also there appear to be some issues that
> have not been fully considered.
> 
> Section 1.2.2
> 
> While the effect of QUIC's response to congestion means that some RTP
> packets will arrive at the receiver later than a user of the RTP flow might
> prefer, it is still preferable to "ceasing transmission" completely until
> the RTP sender has a reason to believe that restarting the flow will not
> result in congestion.¶
> 
> [BA] In contrast to circuit breakers, which do not restrict the ability to
> send RTCP feedback, QUIC congestion control affects RTCP feedback, not just
> RTP. So saying QUIC congestion control is "preferable" seems questionable.
> 
> Moreover, when a single QUIC connection is used to multiplex both RTP-RTCP
> and non-RTP packets as described in Section 1.2.5
> the QUIC connection will still be Internet-safe, with no coordination
> required.
> 
> [BA] While it may be "Internet-safe", delays in RTCP feedback are likely to
> destabilize rate control as well as resulting in challenges to A/V sync.
> So not sure that "Internet-safe" is the only important metric here.
> 
> Section 1.2.3
> 
> One word of caution is in order - RTP implementations may rely on at least
> some minimal periodic RTCP feedback, in order to determine that an RTP flow
> is still active, and is not causing sustained congestion (as described in
> RFC8083, but since this "periodicity" is measured in seconds, the impact of this "duplicate"
> feedback on path bandwidth utilization is likely close to zero.
> 
> [BA] Under congestion, RTCP feedback can potentially be delayed
> substantially. Here the issue is not "bandwidth utilization" but whether
> RTCP receives the transport treatment required for control traffic. Note
> also that similar considerations apply to treatment of audio vs. video.
> Serious problems with a/v sync are possible (or even likely) under
> congestion.

SpencerDawkins:
Also adding explanation of the use of congestion control and rate adaptation terms in this document.

closes #135


SpencerDawkins:
@aboba, could you take a look at the changes I made in the terminology section about "congestion control" and "rate adaptation" term usage? I think what I said is accurate, but I'm also wondering what else we can do.

aboba (commented):
Since QUIC CC is always there, there's no real choice here. You might say that "timely rate adaptation can minimize endpoint queuing delays and loss".

aboba:
Comments provided within the PR. On Oct 20, 2023, at 10:21, Spencer Dawkins ***@***.***> wrote:﻿
@aboba, could you take a look at the changes I made in the terminology section about "congestion control" and "rate adaptation" term usage? I think what I said is accurate, but I'm also wondering what else we can do.

draft-ietf-avtcore-rtp-over-quic.md
```diff
@@ -94,13 +99,21 @@ This is in sharp contrast to "always-on" transport-level encryption in the QUIC
...
 
-In contrast, when RTP is carried over QUIC, QUIC implementations maintain their own estimates of key transport parameters needed to detect and respond to possible congestion, and these are independent of any measurements RTP senders and receivers are maintaining. The result is that even if an RTP sender continues to "send", QUIC congestion avoidance procedures (for example, the procedures defined in {{?RFC9002}}) will cause the RTP packets to be buffered and only placed on the network path as part of a response to detected loss. This happens without any action being requied on the part of RTP senders.
+In contrast, when RTP is carried over QUIC, QUIC implementations maintain their own estimates of key transport parameters needed to detect and respond to possible congestion, and these are independent of any measurements RTP senders and receivers are maintaining. The result is that even if an RTP sender continues to "send", QUIC congestion avoidance procedures (for example, the procedures defined in {{?RFC9002}}) will cause the RTP packets to be buffered and only placed on the network path as part of a response to detected loss. This happens without RTP senders taking any action.
```
mengelbart, line 102: > QUIC congestion avoidance procedures (for example, the procedures defined in {{?RFC9002}}) will cause the RTP packets to be buffered and only placed on the network path as part of a response to detected loss.

I don't understand the sentence why would packets be placed on the network in response to detected loss? Is there a word/fragment of the sentence missing here?

draft-ietf-avtcore-rtp-over-quic.md
```diff
@@ -94,13 +99,21 @@ This is in sharp contrast to "always-on" transport-level encryption in the QUIC
...
+* If the QUIC connection also encapsulates RTCP, this means that these RTCP messages will also be delayed, and will not be sent in a timely manner. This delay can interfere with a sender's ability to stabilize rate control and achieve audio/video synchronization.
 
-While the effect of QUIC's response to congestion means that some RTP packets will arrive at the receiver later than a user of the RTP flow might prefer, it is still preferable to "ceasing transmission" completely until the RTP sender has a reason to believe that restarting the flow will not result in congestion.
+Taken as a whole,
 
-Moreover, when a single QUIC connection is used to multiplex both RTP-RTCP and non-RTP packets as described in {{single-path}}, the QUIC connection will still be Internet-safe, with no coordination required.
+* RTP stream-level rate adaptation can give a better user experience than QUIC connection-level congestion control by minimizing packet loss,
+* but QUIC connection-level congestion control can respond more quickly to the end of congestion than RTP "circuit breakers".
```
mengelbart, line 114: I suggest phrasing this as *QUIC connection-level congestion control __will__ respond more quickly* to emphasize that it is not a choice between the two and that QUIC congestion control is not optional.

draft-ietf-avtcore-rtp-over-quic.md
```diff
@@ -211,6 +224,14 @@ when, and only when, they appear in all capitals, as shown here.
 
 > **Editor's note:** the list of terms below will almost certainly grow in size as the specification matures.
 
+> **Note to the Reader:** the meaning of the terms "congestion control" and "rate adaptation" in the IETF community have evolved over the decades since "slow start" and "congestion avoidance" were added as mandatory to implement in TCP, in {{Section 4.2.2.15 of ?RFC1122}}. At that time, "congestion control" usually referred to "achieving network stability" ({{VJMK88}}), by protecting the network from senders who continue to transmit packets that exceed the ability of the network to carry them, even after packet loss occurs (called "congestion collapse").
+
+> "Rate adaptation" more commonly referred to strategies intended to guide senders on when to send "the next packet", so that one-way delays along the network path remain minimal.
+
+> As more and more general-purpose "congestion control" algorithms focused on avoiding "bufferbloat", as described in {{nested-CC}}, the difference between "congestion control" and "rate adaptation" has blurred in IETF community discussions.
```
mengelbart, line 231: In #128 I removed the {{nested-CC}} section and merged it with the one before because they became very similar. I also removed the note about bufferbloat. Maybe we can directly include that here and remove the reference to the other section?

mengelbart (changes requested):

SpencerDawkins:
@aboba and @mengelbart - I think you both had valuable comments on this text:

> Taken as a whole,
> 
> * RTP stream-level rate adaptation can give a better user experience than QUIC connection-level congestion control by minimizing packet loss,
> * but QUIC connection-level congestion control can respond more quickly to the end of congestion than RTP "circuit breakers".

Just to make sure my change is consistent, I'm checking with both of you about this proposed revision (now committed).

> Taken as a whole,
> 
> * Timely RTP stream-level rate adaptation will give a better user experience by minimizing endpoint queuing delays and packet loss,
> * but in the presence of packet loss, QUIC connection-level congestion control will respond more quickly to the end of congestion than RTP "circuit breakers".

draft-ietf-avtcore-rtp-over-quic.md
SpencerDawkins: @mengelbart - you don't understand 

> I don't understand the sentence why would packets be placed on the network in response to detected loss? Is there a word/fragment of the sentence missing here?

because it's wrong! This should have been something like 

> The result is that even if an RTP sender continues to "send", QUIC congestion avoidance procedures (for example, the procedures defined in {{?RFC9002}}) will cause the RTP packets to be buffered **while QUIC responds to detected packet loss**. This happens without RTP senders taking any action, **but the RTP sender has no control over this QUIC mechanism**. 

Now committed.

draft-ietf-avtcore-rtp-over-quic.md
SpencerDawkins: Hi, @mengelbart - that sounds good, but I'm not sure whether you wanted to fix this, or have me fix it. Since we have a conference call in 9 minutes, let's fix it exactly once!

PR #148, opened 2023-11-07
Update error codes starting from 0x00

mengelbart:
@LPardue noticed we didn't assign 0x00 to any error code. I think it would be nice to have 0x00 as NO_ERROR. I also added the real codes to the error handling section.


SpencerDawkins (approved):
@LPurdue is a smart guy. That's a good thing to do. 

It looks good to merge, to me!

PR #151, opened 2024-01-30
The term is not used in the document

mengelbart:


draft-ietf-avtcore-rtp-over-quic.md
```diff
@@ -266,10 +266,6 @@ Frame:
 Loss-based congestion control algorithm:
 : A congestion control algorithm that uses packet loss, or an Explicit Congestion Notification (ECN) that is interpreted as loss (as in {{?RFC3168}}), as a signal for congestion. Loss-based congestion control algorithms allow senders to send data on a path until packets are dropped by intermediary network elements, which the algorithm treats as a signal of congestion.
 
-Media Encoder:
```
SpencerDawkins, line 269: I think it's correct that we aren't using it - I looked at a couple of other documents (RTP over TCP, etc.) and the focus is encapsulating a media stream that has already been encoded. So, yes, we should remove it!

SpencerDawkins (approved):
You are doing the right thing.

PR #152, opened 2024-01-31
Add considerations for coalescing RTP packets

### LINKED ISSUE #86: Coalescing RTP packets in single QUIC packet
Over UDP, every RTP packet gets sent in a UDP datagram. In QUIC, one could batch multiple small RTP packets together in a QUIC packet. This will probably happen automatically for streams but may depend on some scheduling for datagrams. That scheduling should take real-time requirements into account, i.e., try to optimize by batching, but also not delay packets too much.

Maybe relevant: https://www.rfc-editor.org/rfc/rfc9000#section-13

mengelbart:
close #86

### DIFF
--- a/draft-ietf-avtcore-rtp-over-quic.md
+++ b/draft-ietf-avtcore-rtp-over-quic.md
@@ -1106,6 +1106,40 @@ to determine if 0-RTT data is permissible.
 > the same connection, replayed media packets would be discarded as duplicates by
 > the receiver.
 
+## Coalescing RTP packets in single QUIC packet
+
+Applications have some control over how the QUIC stack maps application data to
+QUIC frames, but applications cannot control how the QUIC stack maps STREAM and
+DATAGRAM frames to QUIC packets {{Section 13 of ?RFC9000}} and {{Section 5 of
+?RFC9308}}.
+
+* When RTP payloads are carried over QUIC streams, the RTP payload is treated as
+  an ordered byte stream that will be carried in QUIC STREAM frames, with no
+  effort to match application data boundaries.
+* When RTP payloads are carried over QUIC DATAGRAMs, each RTP payload data unit
+  is mapped into a QUIC DATAGRAM frame, but
+* QUIC implementations can include multiple STREAM frames from different streams
+  and one or more DATAGRAM frames into a single QUIC packet, and may include
+  other QUIC frames as well.
+
+QUIC stacks are allowed to wait for a short period of time if the queued QUIC
+packet is shorter than the path MTU, in order to optimize for bandwidth
+utilization instead of latency, while real-time applications usually prefer to
+optimize for latency rather than bandwidth utilization. This waiting interval is
+under the QUIC implementation's control, and might be based on knowledge about
+application sending behavior or heuristics to determine whether and for how long
+to wait.
+
+When there are a lot of small DATAGRAM frames (e.g., an audio stream) and a lot
+of large DATAGRAM (e.g., a video stream), it may be a good idea to make sure the
+audio frames can be included in a QUIC packet that also carries video frames
+(i.e., the video frames don't fill the whole QUIC packet). Otherwise, the QUIC
+stack may have to send additional small packets only carrying single audio
+frames, which would waste some bandwidth.
+
+Application designers are advised to take these considerations into account when
+selecting and configuring a QUIC stack for use with RoQ.
+
 # Security Considerations {#sec-considerations}
 
 RoQ is subject to the security considerations of RTP described in


draft-ietf-avtcore-rtp-over-quic.md
```diff
@@ -1110,6 +1110,32 @@ to determine if 0-RTT data is permissible.
...
+
+Applications usually cannot control how the QUIC stack maps application data to
+QUIC packets. While the application controls the content of QUIC DATAGRAM
+frames, stream frames may contain any amount of data. QUIC packets can include
+multiple STREAM frames from different streams and one or more DATAGRAM frames.
+Applications don't necessarily control how the QUIC stack maps STREAM and
+DATAGRAM frames to QUIC packets {{Section 13 of ?RFC9000}} and {{Section 5 of
+?RFC9308}}.
+
+If there is not enough data, QUIC stacks might wait before sending out packets
```
SpencerDawkins, line 1124: For 

> If there is not enough data, QUIC stacks might wait before sending out packets

I'd suggest 

>QUIC stacks are allowed to wait for a short period of time if the queued QUIC packet is shorter than the path MTU, in order

draft-ietf-avtcore-rtp-over-quic.md
```diff
@@ -1110,6 +1110,32 @@ to determine if 0-RTT data is permissible.
...
+to optimize for bandwidth utilization instead of latency, while real-time
+applications usually prefer to optimize for latency rather than bandwidth
+utilization.
```
SpencerDawkins, line 1127: I'd add 

> This waiting interval is under the QUIC implementation's control, and might be based on knowledge about application sending behavior or heuristics to determine whether and for how long to wait.

draft-ietf-avtcore-rtp-over-quic.md
```diff
@@ -1110,6 +1110,32 @@ to determine if 0-RTT data is permissible.
...
+
+When there are a lot of small DATAGRAM frames (e.g., an audio stream) and a lot
+of large DATAGRAM (e.g., a video stream), it may be a good idea to make sure the
+audio frames can be included in a QUIC packet that also carries video frames
+(i.e., the video frames don't fill the whole QUIC packet). Otherwise, the QUIC
+stack may have to send additional small packets only carrying single audio
+frames, which would waste some bandwidth.
```
SpencerDawkins, line 1134: Just to make sure I understand this, is the recommendation that the application queue any related small media objects with an earlier RTP timestamp than the next large media object, before they queue that large media object?

I'm trying to figure out how the application can make sure that the audio frames can be included in a QUIC packet that also carries video frames. I think "sending small objects first" might accomplish that, but whatever we say, we need to give advice that an application developer can implement. 

This only matters if we don't have lots of large objects, right?

draft-ietf-avtcore-rtp-over-quic.md
```diff
@@ -1110,6 +1110,32 @@ to determine if 0-RTT data is permissible.
...
+
+Application designers are advised to take these considerations into account when
+selecting and configuring a QUIC stack for use with RoQ.
```
SpencerDawkins, line 1137: I know we've talked about this before, but some application designers have control over QUIC implementations, and others do not. Is this also a reason why we want people to use "rtp-mux-quic"? If so, we should probably say that, someplace - maybe in a new subsection 7.x. 

(if your QUIC implementation accepts that ALPN string, you ought to be smart about this kind of thing)

SpencerDawkins (approved):
I provided a restructured paragraph, and made a couple of suggestions, and asked a couple of questions, but I know you'll do the right thing.

draft-ietf-avtcore-rtp-over-quic.md
mengelbart: I think any recommendation we can make depends on the level of control the application has over the QUIC stack. Queuing small objects before large ones might work but may require waiting for small objects in the application and possibly delaying large objects. Or delaying small objects to fill the packet and avoid sending small packets. There are some contradictory goals because we'd like to send audio packets as soon as possible but also try to avoid sending small packets, thus having to wait until a packet is _large enough_. It may require some research to give good recommendations here.

draft-ietf-avtcore-rtp-over-quic.md
mengelbart: How is it related to the ALPN? All QUIC implementations I have worked with so far accept any ALPN string I gave them, and I don't think there is any kind of special behavior based on what the ALPN is.

draft-ietf-avtcore-rtp-over-quic.md
```diff
@@ -1110,6 +1110,32 @@ to determine if 0-RTT data is permissible.
...
+## Coalescing RTP packets in single QUIC packet
+
+
+Applications usually cannot control how the QUIC stack maps application data to
+QUIC packets. While the application controls the content of QUIC DATAGRAM
+frames, stream frames may contain any amount of data. QUIC packets can include
+multiple STREAM frames from different streams and one or more DATAGRAM frames.
+Applications don't necessarily control how the QUIC stack maps STREAM and
+DATAGRAM frames to QUIC packets {{Section 13 of ?RFC9000}} and {{Section 5 of
+?RFC9308}}.
```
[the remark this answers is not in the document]
mengelbart: I just copied the whole paragraph :)

draft-ietf-avtcore-rtp-over-quic.md
SpencerDawkins: Yeah, so, that's the thing. I'm not sure whether having our own ALPN string will make any difference at all in practice. One might think that it would make a difference - for instance, a QUIC implementation might focus more on minimizing one-way delays, and not worry as much about probing for bandwidth - but if QUIC implementations don't make that adjustment, I probably need to rethink why we still need "rtp-mux-quic".

Note that this is much more about me needing to think, than us needing to change anything in the spec!

mengelbart: We need "rtp-mux-quic" to tell the application on the other end that we will use the QUIC connection to send RTP packets encapsulated as described in this document.

draft-ietf-avtcore-rtp-over-quic.md
SpencerDawkins: That all makes perfect sense to me. The question is, how the QUIC implementation might adjust its delay timer so that the right thing happens. 

But what I was suggesting, was that if you queue up all the small frames with a timestamp earlier than the next large frame, that should minimize the chance that the QUIC implementation will emit a bunch of small packets. But this is guessing at what the QUIC implementation might do, in the absence of any defined behavior (beyond "I might wait for a short time before sending the packet"), and in the absence of any way to tell the QUIC implementation "if you're waiting a short time for more frames, this stuff is audio, so please set your timer appropriately". 

Maybe THIS is something the "rtp-mux-quic" ALPN could tell the QUIC implementation? 

I love heuristics about how to deal with someone else's heuristics. :grimacing:

draft-ietf-avtcore-rtp-over-quic.md
SpencerDawkins: And now I remember that - thanks for the wake-up call!

PR #156, opened 2024-02-12
Review appendix B

### LINKED ISSUE #126: Review Appendix B for completeness
We have a lot of information in [Appendix B.  Considered RTCP Packet Types and RTP Header Extensions](https://mengelbart.github.io/rtp-over-quic-draft/draft-ietf-avtcore-rtp-over-quic.html#name-list-of-optional-quic-exten), and it looks like we might have some missing table entry details, such as 

> Texas Instruments Extended VoIP Quality Block | | | |

I see that one has a reference in the IANA registry that returns a 404, but we might want to say that in the table, just so readers know we didn't miss anything they need to know. And @joerg-ott is the designated expert for that registry - we should probably ask him how to report the 404, in case he has a better idea where to look than the current reference! :grin:

Also, it would be good for us to point to relevant IANA registries at the beginning of each section (for example, [RTP Control Protocol Extended Reports (RTCP XR) Block Type Registry](https://www.iana.org/assignments/rtcp-xr-block-types/rtcp-xr-block-types.xhtml).

[B.5.1. Compact Header Extensions](https://mengelbart.github.io/rtp-over-quic-draft/draft-ietf-avtcore-rtp-over-quic.html#name-sdes-compact-header-extensi) has several "probably not(?)" entries. Is this just because we haven't looked closely at the 3GPP specifications yet? We should probably look at that - RTP over QUIC should be relevant to 3GPP in the next 5G release or two.

mengelbart:
close #126

### DIFF
--- a/draft-ietf-avtcore-rtp-over-quic.md
+++ b/draft-ietf-avtcore-rtp-over-quic.md
@@ -33,7 +33,7 @@ author:
 informative:
 
   3GPP-TS-26.114:
-    target: https://portal.3gpp.org/desktopmodules/Specifications/specificationId=1404
+    target: https://portal.3gpp.org/desktopmodules/Specifications/SpecificationDetails.aspx?specificationId=1404
     title: "IP Multimedia Subsystem (IMS); Multimedia telephony; Media handling and interaction"
     date: 2023-01-05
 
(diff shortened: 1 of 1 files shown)


SpencerDawkins (approved):
This looks fine to me!

PR #158, opened 2024-02-13
Cleanup of many SHOULDs

### LINKED ISSUE #111: Check SHOULD requirements for ROQ receivers

SpencerDawkins:
closes #111

### DIFF
--- a/draft-ietf-avtcore-rtp-over-quic.md
+++ b/draft-ietf-avtcore-rtp-over-quic.md
@@ -521,12 +521,8 @@ of flows.
 ## QUIC Streams {#quic-streams}
 
 To send RTP/RTCP packets over QUIC streams, a sender MUST open at least one new unidirectional QUIC stream.
-RoQ uses unidirectional streams, because there is no
-synchronous relationship between sent and received RTP/RTCP packets. A peer that
-receives a bidirectional stream with a flow identifier that is associated with
-an RTP or RTCP stream, MUST stop reading from the stream and send a
-STOP\_SENDING frame with the application protocol error code set to
-ROQ\_STREAM\_CREATION\_ERROR.
+RoQ uses unidirectional streams, because there is no synchronous relationship between sent and received RTP/RTCP packets.
+A peer that receives a bidirectional stream with a flow identifier that is associated with an RTP or RTCP stream, MUST stop reading from the stream and send a CONNECTION_CLOSE frame with the frame type set to APPLICATION_ERROR and the error code set to ROQ_STREAM_CREATION_ERROR.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-avtcore-rtp-over-quic.md
```diff
@@ -526,11 +524,11 @@ To send RTP/RTCP packets over QUIC streams, a sender MUST open at least one new
 RoQ uses unidirectional streams, because there is no
 synchronous relationship between sent and received RTP/RTCP packets. A peer that
 receives a bidirectional stream with a flow identifier that is associated with
-an RTP or RTCP stream, SHOULD stop reading from the stream and send a
+an RTP or RTCP stream, MUST stop reading from the stream and send a
```
mengelbart, line 527: I wonder if this should lead to closing the connection because it is a protocol violation? We make that change in a different PR though.

mengelbart (approved):
Thanks, I think this is ready to be merged.

draft-ietf-avtcore-rtp-over-quic.md
SpencerDawkins: Actually, you're right - we talked about that, but somehow it didn't make the transition into our shared notes from our call. Let's wait to merge until I can update the PR (I'm having odd Windows problems right now, so that may take a day or two).

SpencerDawkins: @mengelbart - I think I got this one right. Please let me know! 

If I did, we should be able to merge the PR.

PR #162, opened 2024-02-17
Move details about ECN and L4S into Section 7.1

### LINKED ISSUE #160: Moving details about ECN and L4S from terminology to the right place in Section 7 (probably 7.1)
There are WAY too many details to be in a terminology section, especially when the draft already has a section where they are more appropriate!

SpencerDawkins:
closes #160


mengelbart:
This looks good to me.

PR #164, opened 2024-02-18
Add references for actual IANA registries in Appendix B

### LINKED ISSUE #157: Add references for each IANA reference named in the document
@mengelbart noticed that the link in our document for DBI is broken, and @SpencerDawkins noticed that we don't seem to have references for each IANA registry that we name in the document. It would be good to add them, and clean up the DBI reference.

SpencerDawkins:
Note that I also changed subsection titles in this document to match the titles used on each IANA registry.

closes #157 .


mengelbart:
Looks good, I think we can merge this.

PR #165, opened 2024-02-18
Convert ambiguous "datagrams" and lower-case "QUIC datagrams" to QUIC DATAGRAMs.

### LINKED ISSUE #147: Make sure "UDP datagram" and "QUIC datagram" are used consistently
We say that "datagram" has two meanings in the terminology section, and say that "datagram" without qualification refers to QUIC datagrams, but if we always qualify the term, that would be easier for the reader.

SpencerDawkins:
Note that we said "datagrams are QUIC datagrams", but a better way to handle this is to use upper-case "DATAGRAM" to mean QUIC DATAGRAMs.

I didn't convert "QUIC DATAGRAMs" to something like "DATAGRAM frames", but that would be easy enough to do (now that I've removed most/all of the line breaks in the middle of sentences about DATAGRAMs). 

I note that there were some uses of terms like "RoQ datagrams". Please let me know if you don't think I handled them correctly. 

closes #147

### DIFF
--- a/draft-ietf-avtcore-rtp-over-quic.md
+++ b/draft-ietf-avtcore-rtp-over-quic.md
@@ -78,7 +78,7 @@ This specification describes an application usage of QUIC ({{?RFC9308}}).
 As a baseline, the specification does not expect more than a standard QUIC implementation
 as defined in {{!RFC8999}}, {{!RFC9000}}, {{!RFC9001}}, and {{!RFC9002}},
 providing a secure end-to-end transport that is also expected to work well through NATs and firewalls.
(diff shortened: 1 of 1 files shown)


mengelbart:
I think we can merge this. There are a few more cases of _RoQ datagram_, should we keep them or change them too, either in this or another PR?

SpencerDawkins:
> I think we can merge this. There are a few more cases of _RoQ datagram_, should we keep them or change them too, either in this or another PR?

I think we should keep them, but I want to mention _RoQ datagram_ and _RoQ stream_ in the terminology section. Stay tuned, of course.

mengelbart (approved):

PR #166, opened 2024-02-19
Add guidance on choosing streams and datagrams

### LINKED ISSUE #114: Guidance on RoQ DATAGRAM use
@SpencerDawkins and @mengelbart sait through some really interesting discussion at WEBTRAN about pooled connections ...

SpencerDawkins:
closes #114


SpencerDawkins:
@mengelbart - I think this is ready to review. My biggest question is "did I put this in the right place?".

draft-ietf-avtcore-rtp-over-quic.md
```diff
@@ -856,6 +856,20 @@ in the shared QUIC connection. For this reason, applications multiplexing
 multiple streams in one connection SHOULD implement some form of stream
 prioritization or bandwidth allocation.
 
+# Guidance on Choosing QUIC Streams, QUIC DATAGRAMs, or a Mixture {#s-d-m-guidance}
+
+As noted in {{streams-and-datagrams}}, this specification does not take a position on using QUIC streams, QUIC DATAGRAMs, or some mixture of both, for any particular RoQ use case or application. It does seem useful to include observations that might guide implemeters who will need to make choices about that.
```
mengelbart, line 861:
```suggestion
As noted in {{streams-and-datagrams}}, this specification does not take a position on using QUIC streams, QUIC DATAGRAMs, or some mixture of both, for any particular RoQ use case or application. It does seem useful to include observations that might guide implementers who will need to make choices about that.
```

draft-ietf-avtcore-rtp-over-quic.md
```diff
@@ -856,6 +856,20 @@ in the shared QUIC connection. For this reason, applications multiplexing
...
+
+One implementation goal might be to minimize processing overhead, for applications that are migrating from RTP over UDP to RoQ. These applications don't rely on any transport protocol behaviors beyond UDP, which can be described as "nothing beyond IP, except but multiplexing". They might be motivated by one or more of the advantages of encapsulating RTP in QUIC that are described in {{motivations}}, but they do not need any of the advantages that would apply when encapsulating RTP in QUIC streams. For these applications, simply placing each RTP packet in a QUIC DATAGRAM frame when it becomes available would be sufficient, with no QUIC streams at all.
```
mengelbart, line 863:
```suggestion
One implementation goal might be to minimize processing overhead, for applications that are migrating from RTP over UDP to RoQ. These applications don't rely on any transport protocol behaviors beyond UDP, which can be described as "nothing beyond IP, except multiplexing". They might be motivated by one or more of the advantages of encapsulating RTP in QUIC that are described in {{motivations}}, but they do not need any of the advantages that would apply when encapsulating RTP in QUIC streams. For these applications, simply placing each RTP packet in a QUIC DATAGRAM frame when it becomes available would be sufficient, with no QUIC streams at all.
```

draft-ietf-avtcore-rtp-over-quic.md
```diff
@@ -856,6 +856,20 @@ in the shared QUIC connection. For this reason, applications multiplexing
...
+
+Some applications might have implementation goals that don't fit easily into "QUIC streams only" or "QUIC DATAGRAMs only" categories. For example, another implementation goal might be to use QUIC streams to carry RTP video frames, but to use QUIC DATAGRAMs to carry RTP audio frames, which are typically much smaller. Because humans tend to tolerate inconsistent behavior in video better than inconsistent behavior in audio, the application might add Forward Error Correction {{!RFC6363}} to audio samples and encapsulate the result in QUIC DATAGRAMs while encapsulating RTP video packets in QUIC streams.
+
+As noted in {{multiplexing}}, all RoQ streams and datagrams begin with a flow identifier. This allows a RoQ sender to begin by encapulating related RTP packets in a stream and then switch to carrying them in QUIC DATAGRAMs, or vice versa. RoQ receivers need to be prepared to accept any valid RTP packet with a given flow identifier, whether it started by being encapsulated in QUIC streams or in QUIC DATAGRAMs, and RoQ receivers need to be prepared to accept RTP flows that switch from QUIC stream encapsulation to QUIC DATAGRAMs, or vice versa.
```
mengelbart, line 869:
```suggestion
As noted in {{multiplexing}}, all RoQ streams and datagrams begin with a flow identifier. This allows a RoQ sender to begin by encapsulating related RTP packets in a stream and then switch to carrying them in QUIC DATAGRAMs, or vice versa. RoQ receivers need to be prepared to accept any valid RTP packet with a given flow identifier, whether it started by being encapsulated in QUIC streams or in QUIC DATAGRAMs, and RoQ receivers need to be prepared to accept RTP flows that switch from QUIC stream encapsulation to QUIC DATAGRAMs, or vice versa.
```

draft-ietf-avtcore-rtp-over-quic.md
```diff
@@ -856,6 +856,20 @@ in the shared QUIC connection. For this reason, applications multiplexing
...
+
+Because QUIC provides a capability to migrate connections for various reasons, including recovering from a path failure ({{Section 9 of !RFC9000}}), a RoQ sender has the opportunity to revisit decisions about which RTP packets are encapsulated in QUIC streams, and which RTP packets are encapsulated in QUIC DATAGRAMs, when a QUIC connection migrates. Again, RoQ receivers need to be prepated for this eventuality.
```
mengelbart, line 871: I think we don't need this paragraph. As noted before, a sender always has the option to switch between streams and datagrams, not only when connection migration happens. Also, a receiver must also always be prepared for receiving RTP in streams or datagrams.

mengelbart (commented):
I think it should be combined with section 3.1. We could move it somewhere else, but I think the text of what is now 3.1 should be close to the text added in this PR.

draft-ietf-avtcore-rtp-over-quic.md
SpencerDawkins: That's ... fair. I added this paragraph, and then realized that the sender was always in charge, not only during connection migration, but thought that distinguishing the two cases was useful. Actually, it's not. 	:thinking:

PR #169, opened 2024-02-20
Clean up some terminology that is not actually used

mengelbart:
*Delay-based* is not used and *Low-latency congestion control algorithm* is used only once. This commit moves the explanation to where it is used to avoid confusion, where a reader did not see the definition, because it is hidden in the terminology section.

*Loss-based* is not used in the document.

*QUIC congestion controller* is not used in the document.

*RTP congestion controller* is not used in the document.

It was helpful to define these terms when writing the document, but I think listing them in the terminology section without using them may be more confusing than helpful.

### DIFF
--- a/draft-ietf-avtcore-rtp-over-quic.md
+++ b/draft-ietf-avtcore-rtp-over-quic.md
@@ -281,29 +281,17 @@ Datagram:
 differently, the term datagram in this document refers to a QUIC Datagram as defined in
 {{!RFC9221}}.
 
-Delay-based or Low-latency congestion control algorithm:
-: A congestion control algorithm that aims at keeping queues, and thus the latency, at intermediary network elements as short as possible. Delay-based congestion control algorithms use, for example, an increasing one-way delay as a signal of impending congestion, and adjust the sending rate to prevent continued increases in one-way delay.
-
 Endpoint:
 : A QUIC server or client that participates in an RoQ session.
 
 Frame:
 : A QUIC frame as defined in {{!RFC9000}}.
 
-Loss-based congestion control algorithm:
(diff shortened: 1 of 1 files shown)


draft-ietf-avtcore-rtp-over-quic.md
```diff
@@ -772,8 +760,13 @@ free to choose any congestion control algorithm as long as they follow the
 guidelines specified in {{Section 3 of ?RFC8085}}, and QUIC implementors make
 use of this freedom.
 
-For real-time traffic, it is best that the QUIC implementation use a congestion controller that
-seeks to minimize queueing delays.
+For real-time traffic, it is best that the QUIC implementation use a congestion
+controller that aims at keeping queues, and thus the latency, at intermediary
+network elements as short as possible. Delay-based congestion control algorithms
+use, for example, an increasing one-way delay as a signal of impending
```
SpencerDawkins, line 766: I'd suggest "might use, for example", because I don't know that's true for all of them, but otherwise, this looks lovely. 

One minor point - you said "hidden in the terminology section", and IIRC, I started adding subsections as part of the Introduction, before I realized that (in total) the terminology section would appear after NINE subsections. Does it make sense to you to move Section 1.2 ("Motivations") into its own top-level section, following the terminology section?

SpencerDawkins (approved):

draft-ietf-avtcore-rtp-over-quic.md
mengelbart: Thanks, I updated the PR.

Yes, we could do that. But let's make a separate PR for that. I can prepare it.

PR #174, opened 2024-02-20
Remove non-RTP/RTCP protocol multiplexing

mengelbart:
As discussed at the last interim meeting, this proposal removes multiplexing with non-RTP/RTCP protocols. The same features could be implemented by defining new Multiplexing schemes and respective ALPNs in other documents. That would have the advantage of not requiring a new protocol negotiation mechanism at a higher layer. It would also be easier to define multiplexing for known protocols rather than providing a general solution that works for every protocol that one might want to use.

### DIFF
--- a/draft-ietf-avtcore-rtp-over-quic.md
+++ b/draft-ietf-avtcore-rtp-over-quic.md
@@ -151,7 +151,13 @@ This is especially useful in certain conferencing topologies, where otherwise se
 
 ### Multiplexing RTP, RTCP, and Non-RTP Flows on a Single QUIC Connection {#single-path}
 
-In order to conserve ports, especially at NATs and Firewalls, this specification defines a flow identifier, so that multiple RTP flows, RTCP flows, and non-RTP flows can be distinguished even if they are carried on the same QUIC connection. This is described in more detail in {{multiplexing}}.
+This specification defines a flow identifier for multiplexing multiple RTP and
+RTCP ports on the same QUIC connection to conserve ports, especially at NATs and
+Firewalls. {{multiplexing}} describes the multiplexing in more detail. Future
+extensions could further build on the flow identifier to multiplex RTP/RTCP with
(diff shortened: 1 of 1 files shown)


draft-ietf-avtcore-rtp-over-quic.md
```diff
@@ -151,7 +151,11 @@ This is especially useful in certain conferencing topologies, where otherwise se
 
 ### Multiplexing RTP, RTCP, and Non-RTP Flows on a Single QUIC Connection {#single-path}
 
-In order to conserve ports, especially at NATs and Firewalls, this specification defines a flow identifier, so that multiple RTP flows, RTCP flows, and non-RTP flows can be distinguished even if they are carried on the same QUIC connection. This is described in more detail in {{multiplexing}}.
+This specification defines a flow identifier for multiplexing multiple RTP and
+RTCP ports on the same QUIC connection to conserve ports, especially at NATs and
+Firewalls. {{multiplexing}} describes the multiplexing in more detail. Future
+extensions could further build on the flow identifier to multiplex RTP/RTCP with
+other protocols on the same connection.
```
SpencerDawkins, line 158: > Future extensions could further build on the flow identifier to multiplex RTP/RTCP with other protocols on the same connection.

I think it's worth pointing out (as we have realized while working on this specification) that there's more to multiplexing than just disambiguating and demultiplexing. Perhaps 

> Future extensions could further build on the flow identifier to multiplex RTP/RTCP with other protocols on the same connection, **as long as these protocols can co-exist with RTP/RTCP without interfering with the ability of this connection to carry real-time media.** 

Does that make sense?

SpencerDawkins (approved):
I left one comment ("Do The Right Thing"), but this is ready to merge without changes.

draft-ietf-avtcore-rtp-over-quic.md
mengelbart: Makes sense. I updated the PR.

PR #178, opened 2024-02-29
Add Futures section and change document category to "exp" (Experimental)

### LINKED ISSUE #161: Adding descriptions of experiments and future work for publication as Experimental
@aboba suggested that we add this issue, at the [interim-2024-avtcore-01](https://datatracker.ietf.org/meeting/interim-2024-avtcore-01/session/avtcore) meeting, Various points described here might end up in the shepherd write-up, a "future work" section in the document, or both. 

Better understanding of useful guidelines and recommendations for congestion control and rate adaptation for multiple RoQ flows (whether streams or datagrams). 

Guidance for using RoQ over QUIC multipath. 

Coexistence between RoQ on its own ALPN and QUIC over ICE, if the QUIC working group produces a "QUIC with ICE" specification. 

Stable reference for implementers to interop test with, and report results.

SpencerDawkins:
Closes #161

### DIFF
--- a/draft-ietf-avtcore-rtp-over-quic.md
+++ b/draft-ietf-avtcore-rtp-over-quic.md
@@ -1131,27 +1131,27 @@ selecting and configuring a QUIC stack for use with RoQ.
 
 This specification represents considerable work and discussion within the IETF, and describes RoQ in sufficient detail that an implementer can build a RoQ application, but we recognize that additional work is likely, after we have sufficient experience with RoQ to guide that work. Possible directions would include
 
-* Better guidance on transport for RTCP (for example, when to use QUIC streams vs. QUIC datagrams for RTCP transport).
+* Better guidance on transport for RTCP (for example, when to use QUIC streams vs. QUIC datagrams).
 
 * Better guidance on the use of realtime-friendly congestion control algorithms (for example, Copa {{Copa}}, L4S {{?RFC9330}}, etc.).
 
-* Better guidelines and recommendations for congestion control and rate adaptation for multiple RoQ flows (whether streams or datagrams).
+* Better guidance for congestion control and rate adaptation for multiple RoQ flows (whether streams or datagrams).
 
-* Possible guidelines for connection sharing between RoQ and non-RoQ flows.
+* Possible guidance for connection sharing between RoQ and non-RoQ flows, including considerations for congestion control and rate adaptation, scheduling, prioritization, and which ALPNs to use.
 
 For these reasons, publication of this specification as a stable reference for implementers to test with, and report results, seems useful.
 
-In addition, as noted in {{new-quic}}, one of the motivations for using QUIC as a transport for RTP is to exploit new QUIC extensions as they become available. We can anticipate additional work if these proposed QUIC extensions become available:
+In addition, as noted in {{new-quic}}, one of the motivations for using QUIC as a transport for RTP is to exploit new QUIC extensions as they become available. We noted several proposed QUIC extensions in {{optional-extensions}}, but these proposals are all solving relevant problems, and those problems are worthy of attention, no matter how they are solved for the QUIC protocol.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-avtcore-rtp-over-quic.md
```diff
@@ -1122,6 +1127,32 @@ frames, which would waste some bandwidth.
...
+
+This specification represents considerable work and discussion within the IETF, and describes RoQ in sufficient detail that an implementer can build a RoQ application, but we recognize that additional work is likely, after we have sufficient experience with RoQ to guide that work. Possible directions would include
+
+* Better guidance on transport for RTCP (for example, when to use QUIC streams vs. QUIC datagrams for RTCP transport).
+
+* Better guidance on the use of realtime-friendly congestion control algorithms (for example, Copa {{Copa}}, L4S {{?RFC9330}}, etc.).
+
+* Better guidelines and recommendations for congestion control and rate adaptation for multiple RoQ flows (whether streams or datagrams).
+
+* Possible guidelines for connection sharing between RoQ and non-RoQ flows.
```
mengelbart, line 1140:
```suggestion
* Possible guidelines for connection sharing between RoQ and non-RoQ flows, including guidance on the use of congestion control, scheduling, prioritization, and guidance on which ALPNs to use.
```

draft-ietf-avtcore-rtp-over-quic.md
```diff
@@ -1122,6 +1127,32 @@ frames, which would waste some bandwidth.
...
+
+For these reasons, publication of this specification as a stable reference for implementers to test with, and report results, seems useful.
+
+In addition, as noted in {{new-quic}}, one of the motivations for using QUIC as a transport for RTP is to exploit new QUIC extensions as they become available. We can anticipate additional work if these proposed QUIC extensions become available:
+
+* Guidance for using RoQ over the Multipath Extension for QUIC {{?I-D.draft-ietf-quic-multipath}}.
+
+* Guidance for using RoQ with QUIC extensions mentioned in {{optional-extensions}} for using Interactive Connectivity Establishment (ICE) {{?RFC8445}} with QUIC, including guidance for coexistence between QUIC connections established using ALPNs and QUIC connections established using ICE.
```
mengelbart, line 1148: I would split this and move the part about ALPN up (see comment above). QUIC needs an ALPN, no matter whether the connection is established via ICE or not. And that ALPN could be `roq` or something else, as long as the document that specifies the ALPN clearly defines the application protocol.

```suggestion
* Guidance for using RoQ with NAT traversal solutions. This could use Interactive Connectivity Establishment (ICE) {{?RFC8445}} with QUIC or other NAT traversal solutions, e.g., using potential QUIC extensions.
```

SpencerDawkins: This is EXACTLY right. I'll look at the other possible topics, to see if they could/should be phrased more generally (we don't want to use specific individual drafts, we want to solve the problems those drafts are intended to solve - duh!).

draft-ietf-avtcore-rtp-over-quic.md
SpencerDawkins: This is good in theory, but in practice, 

> * Possible guidelines for connection sharing between RoQ and non-RoQ flows, including guidance on the use of congestion control, scheduling, prioritization, and guidance on which ALPNs to use.

overlaps with 

> * Possible guidance for connection sharing between RoQ and non-RoQ flows.

and 

> * Guidance for using RoQ with QUIC extensions mentioned in {{optional-extensions}} for using Interactive Connectivity Establishment (ICE) {{?RFC8445}} with QUIC, including guidance for coexistence between QUIC connections established using ALPNs and QUIC connections established using ICE.

I'm going to use almost every word you suggested - I'll propose using them in different places. :upside_down_face:

(which OUGHT to include at least congestion control, scheduling, prioritization).

I'll also use "guidance" consistently (rather than a mixture of "guidance", "guidelines and recommendations", and "guidelines", which I intended to all mean the same thing).

PR #184, opened 2024-03-01
Describe considerations for early data with 0-RTT

### LINKED ISSUE #181: Specify how the contents of the 0-RTT initial packet payload are handled in RoQ
This Editor's Note

> Editor's Note: Since the QUIC connection will often be created in the context of an existing signaling relationship (e.g., using WebRTC or SIP), specific 0-RTT keying material could be exchanged to prevent replays across sessions. Within the same connection, replayed media packets would be discarded as duplicates by the receiver.

is worth its own issue, because [Section 9.2 in RFC 9001](https://www.rfc-editor.org/rfc/rfc9001#section-9.2) says

> Application protocols MUST either prohibit the use of extensions that carry application semantics in 0-RTT or provide replay mitigation strategies.

and, peeking at [Section 10 9 of RFC 9114](https://datatracker.ietf.org/doc/html/rfc9114#name-early-data), HTTP/3 does replay protection this way 

> The use of 0-RTT with HTTP/3 creates an exposure to replay attack. The anti-replay mitigations in [[HTTP-REPLAY](https://datatracker.ietf.org/doc/html/rfc8470)] MUST be applied when using HTTP/3 with 0-RTT.

and none of the mitigations in [Section 3 of HTTP-REPLAY (RFC 8470)](https://datatracker.ietf.org/doc/html/rfc8470#autoid-4) seem applicable. 

We THINK the right answer is to allow 0-RTT connections, but to require RoQ endpoints to ignore any packet payload in an 0-RTT initial packet received during the TLS handshake,

SpencerDawkins:
closes #181

### DIFF
--- a/draft-ietf-avtcore-rtp-over-quic.md
+++ b/draft-ietf-avtcore-rtp-over-quic.md
@@ -1064,26 +1064,30 @@ such as retransmissions) need to be prepared to cope with such spikes.
 > **Editor's Note:** It may be desirable that the API provides an indication
 > of connection migration event for either case.
 
-## 0-RTT considerations
+## 0-RTT and Early Data considerations
 
 For repeated connections between peers, the initiator of a QUIC connection can
-use 0-RTT data for both QUIC STREAM frames and DATAGRAMs. As such packets are subject to
-replay attacks, applications shall carefully specify which data types and operations
-are allowed.  0-RTT data may be beneficial for use with RoQ to reduce the
-risk of media clipping, e.g., at the beginning of a conversation.
+use 0-RTT initial packets to establish a connection that will be used for QUIC STREAM frames, DATAGRAMs, or both.
+As such packets are subject to replay attacks, RoQ applications MUST carefully specify which data types and operations
+are allowed.
 
-This specification defines carrying RTP on top of QUIC and thus does not finally
-define what the actual application data are going to be.  RTP typically carries
-ephemeral media contents that is rendered and possibly recorded but otherwise
-causes no side effects. Moreover, the amount of data that can be carried as 0-RTT
-data is rather limited.  But it is the responsibility of the respective application
-to determine if 0-RTT data is permissible.
+{{Section 9.2 of !RFC9001}} says
 
-> **Editor's Note:** Since the QUIC connection will often be created in the context
-> of an existing signaling relationship (e.g., using WebRTC or SIP), specific 0-RTT
-> keying material could be exchanged to prevent replays across sessions.  Within
-> the same connection, replayed media packets would be discarded as duplicates by
-> the receiver.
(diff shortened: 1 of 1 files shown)


draft-ietf-avtcore-rtp-over-quic.md
```diff
@@ -1064,26 +1064,30 @@ such as retransmissions) need to be prepared to cope with such spikes.
...
+A risk-averse RoQ implementer can simply ignore any RTP packets carried in the 0-RTT Initial packet payload in their RoQ implementation, reasoning that RTP applications are resilient to RTP packet loss.
+
+A risk-taking RoQ implementer can simply accept any RTP packets carried in the 0-RTT Initial packet payload in their RoQ implementation,
+reasoning that
+
+- RTP typically carries ephemeral media contents that is rendered and possibly recorded but otherwise causes no side effects,
+
+- the amount of data that can be carried as packet payload in an 0-RTT Initial packet is rather limited, and
+
+- RTP implementations are likely to discard any replayed media packets as duplicates.
```
mengelbart, line 1090: Should we add/keep the note that specific 0-RTT keying material to prevent replays across sessions could be exchanged in out-of-band signaling?

mengelbart (approved):

draft-ietf-avtcore-rtp-over-quic.md
SpencerDawkins: That sounds reasonable to me. One additional question - are there other concerns about using 0-RTT, beyond the risk of accepting early data from replayed QUIC packets?

mengelbart: I think another concern is that the endpoints may need to remember application-specific data. But I don't think there is anything RoQ-specific except the usual signaling for RTP session setup, which is not part of the QUIC connection setup.

SpencerDawkins:
Looking at [RFC 9000 Section 6.1](https://datatracker.ietf.org/doc/html/rfc9000#name-sending-version-negotiation), we think QUIC endpoints should ("eventually") stop sending 0-RTT packets and send Initial packets instead, but "eventually" is not a great plan for RTP. We should mention this in the RoQ specification. 

Looking [RFC 9001 section 4.6.1](https://datatracker.ietf.org/doc/html/rfc9001#name-enabling-0-rtt), we think that the way a QUIC endpoint knows that the other QUIC endpoint is prepared to do 0-RTT is by the presence of a max_early_data_size parameter with the sentinel value, or when the 0-RTT recipient rejects the 0-RTT packet as described in RFC 9001 Section 4.6.1.

SpencerDawkins:
We can also add a note for implementers that we aren't defining any new transport parameterrs that a QUIC implementation should remember from previous connections, so that adding connections between two endpoints with the same connection ID can be invisible to the RoQ application. If additional out of band signaling happens, that needs to be visible to the RoQ application.

PR #185, opened 2024-03-12
Fix some spelling mistakes

mengelbart:
I chose American spelling over British (e.g., *acknowledgments* over *acknowledgements*) to be consistent, except for *cancellation* and *cancelled*, because that would mean we have to change error code names. We may still want to do that, but better in a separate PR to keep this one editorial only.


draft-ietf-avtcore-rtp-over-quic.md
```diff
@@ -1434,8 +1434,8 @@ The IANA registry for this section is {{IANA-RTP-CHE}}.
 | urn:ietf:params:rtp-hdrext:csrc-audio-level | Mixer-to-client audio level indicators | {{?RFC6465}} | no |
 | urn:3gpp:video-orientation:6 | Higher granularity (6-bit) coordination of video orientation (CVO) feature, see clause 6.2.3 | {{3GPP-TS-26.114}} | probably not(?) |
 | urn:3gpp:video-orientation | Coordination of video orientation (CVO) feature, see clause 6.2.3 | {{3GPP-TS-26.114}} | probably not(?) |
-| urn:3gpp:roi-sent | Signalling of the arbitrary region-of-interest (ROI) information for the sent video, see clause 6.2.3.4 | {{3GPP-TS-26.114}} | probably not(?) |
-| urn:3gpp:predefined-roi-sent | Signalling of the predefined region-of-interest (ROI) information for the sent video, see clause 6.2.3.4 | {{3GPP-TS-26.114}} | probably not(?) |
+| urn:3gpp:roi-sent | Signaling of the arbitrary region-of-interest (ROI) information for the sent video, see clause 6.2.3.4 | {{3GPP-TS-26.114}} | probably not(?) |
+| urn:3gpp:predefined-roi-sent | Signaling of the predefined region-of-interest (ROI) information for the sent video, see clause 6.2.3.4 | {{3GPP-TS-26.114}} | probably not(?) |
```
SpencerDawkins, line 1438: For these 3GPP URNs, we might or might not want to change the spellings, because I think these match [entries in 3GPP documents](https://www.3gpp.org/3gpp-groups/core-network-terminals-ct/ct-wg1/uniform-resource-identifier-uri-list) that spell it as "signalling". I don't know if that really matters, and if it does, the RFC Editor is going to be the arbiter of taste anyway, but I should ask while I'm reviewing the PR, right?

SpencerDawkins (approved):
For the question about American vs. British spelling of "acknowledgment", for what it's worth, we have a section title in the document that uses the American spelling. If we're not matching something else (like an RFC that uses British spellings), I think the American spelling is a good place to start!

draft-ietf-avtcore-rtp-over-quic.md
mengelbart: Good point, I reverted these and also the *Generic negative acknowledgement* above.

SpencerDawkins: Thanks! It really does take teamwork ...

PR #190, opened 2024-04-02
Clarify usage of "RTP" and "RTCP" in the document

### LINKED ISSUE #188: Clarify RTP vs RTCP vs RTP/RTCP handling early in the document
RTP and RTCP were originally defined in the same specification ([RFC 3550](https://datatracker.ietf.org/doc/html/rfc3550)), and later specifications such as [RFC 5761](https://datatracker.ietf.org/doc/html/rfc5761) allowed RTP and RTCP to be multiplexed on the same transport port. The RTP over QUIC specification supports this multiplexing, but at least @SpencerDawkins has not been careful to distinguish between RTP, RTCP, and multiplexed RTP/RTCP in the text. Context clues help the reader, but clear text would help even more. 

For example, the document title, "RTP over QUIC", could just as well be "RTP/RTCP over QUIC".

SpencerDawkins:
closes #188

### DIFF
--- a/draft-ietf-avtcore-rtp-over-quic.md
+++ b/draft-ietf-avtcore-rtp-over-quic.md
@@ -1,5 +1,5 @@
 ---
-title: "RTP and RTCP over QUIC (RoQ)"
+title: "RTP over QUIC (RoQ)"
 docname: draft-ietf-avtcore-rtp-over-quic-latest
 category: exp
 date: {DATE}
@@ -106,7 +106,7 @@ with DTLS on top to secure the media exchange and occasionally TCP (and possibly
 TLS) as a fallback.
 
 This document describes an application usage of QUIC ({{?RFC9308}}).
-As a baseline, the Document does not expect more than a standard QUIC implementation
+As a baseline, the document does not expect more than a standard QUIC implementation
 as defined in {{!RFC8999}}, {{!RFC9000}}, {{!RFC9001}}, and {{!RFC9002}},
 providing a secure end-to-end transport that is also expected to work well through NATs and firewalls.
 Beyond this baseline, real-time applications can benefit from QUIC extensions such as unreliable DATAGRAMs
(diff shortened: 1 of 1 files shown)


draft-ietf-avtcore-rtp-over-quic.md
```diff
@@ -103,23 +105,18 @@ packet losses, the default underlying transport protocol has been UDP, recently
 with DTLS on top to secure the media exchange and occasionally TCP (and possibly
 TLS) as a fallback.
 
-This specification describes an application usage of QUIC ({{?RFC9308}}).
-As a baseline, the specification does not expect more than a standard QUIC implementation
+This document describes an application usage of QUIC ({{?RFC9308}}).
+As a baseline, the Document does not expect more than a standard QUIC implementation
```
mengelbart, line 109:
```suggestion
As a baseline, the document does not expect more than a standard QUIC implementation
```

draft-ietf-avtcore-rtp-over-quic.md
```diff
@@ -103,23 +105,18 @@ packet losses, the default underlying transport protocol has been UDP, recently
...
 as defined in {{!RFC8999}}, {{!RFC9000}}, {{!RFC9001}}, and {{!RFC9002}},
 providing a secure end-to-end transport that is also expected to work well through NATs and firewalls.
 Beyond this baseline, real-time applications can benefit from QUIC extensions such as unreliable DATAGRAMs
 {{!RFC9221}}, which provides additional desirable properties for
 real-time traffic (e.g., no unnecessary retransmissions, avoiding head-of-line
 blocking).
 
-## What's in Scope for this Specification {#in-scope}
+## What's in Scope for this document {#in-scope}
```
mengelbart, line 117:
```suggestion
## What's in Scope for this Document {#in-scope}
```

draft-ietf-avtcore-rtp-over-quic.md
```diff
@@ -1,5 +1,5 @@
 ---
-title: "RTP over QUIC (RoQ)"
+title: "RTP and RTCP over QUIC (RoQ)"
```
mengelbart, line 2: I agree with the other changes, but I don't think it is necessary to change the title. RTCP is part of RTP in RFC 3550, which only has RTP in its title.

SpencerDawkins: > I agree with the other changes, but I don't think it is necessary to change the title. RTCP is part of RTP in RFC 3550, which only has RTP in its title.

@mengelbart - thank you for pushing back on this change. My intention going into this PR was to make it obvious that what we said about encapsulation applied to both RTP Data Transfer Protocol and RTP Control Protocol, but after I added the Note to the Reader about this, I ought to have reversed the proposed change to the title. 

And I am slightly cursing under my breath that "RTCP" caught on, but "RTDTP" never did! :face_with_spiral_eyes:

PR #198, opened 2024-04-13
Add pacing recommendation reference to RFC9002

mengelbart:


draft-ietf-avtcore-rtp-over-quic.md
```diff
@@ -758,7 +758,7 @@ to determine when a packet can be sent to avoid bursts and minimize variation in
 inter-packet arrival times. The currently proposed congestion control algorithms
 for real-time communications (e.g., SCReAM and NADA) provide such pacing
 mechanisms, and QUIC recommends pacing for senders based on the congestion
-control algorithm.
+control algorithm ({{Section 7.7 of !RFC9002}}).
```
SpencerDawkins, line 761: I don't think this is quite right, because RFC9002 is only one congestion control algorithm (that's why I didn't look there, for a reference for what QUIC recommends :upside_down_face:) - RFC9000 calls {{Section 7.7 of !RFC9002}} an exemplary congestion control algorithm. 

I think it is good to add a reference to RFC9002 here, but maybe something like 

> The currently proposed congestion control algorithms for real-time communications (e.g., SCReAM and NADA) provide such pacing mechanisms, and the QUIC exemplary congestion control algorithm ({{Section 7.7 of !RFC9002}}) recommends pacing for senders.

SpencerDawkins (approved):

draft-ietf-avtcore-rtp-over-quic.md
mengelbart: Sounds good, I'll update it!

PR #205, opened 2024-04-22
Replace DTLS with SRTP

mengelbart:
From [RFC 9147](https://www.rfc-editor.org/rfc/rfc9147#section-3)

> telephony utilizes DTLS for key establishment and the Secure Real-time Transport Protocol (SRTP) for protection of data

I think it is better to reference SRTP here.

### DIFF
--- a/draft-ietf-avtcore-rtp-over-quic.md
+++ b/draft-ietf-avtcore-rtp-over-quic.md
@@ -101,10 +101,13 @@ adaptation for RTP without relying on RTCP feedback.
 The Real-time Transport Protocol (RTP) {{!RFC3550}} is generally used to carry
 real-time media for conversational media sessions, such as video conferences,
 across the Internet.  Since RTP requires real-time delivery and is tolerant to
-packet losses, the default underlying transport protocol has historically been UDP {{?RFC0768}}. More recently,
-DTLS {{?RFC9147}} has been used as an underlying transport in order to secure the media exchange,
-and other transport protocols such as TCP {{?RFC9293}} and TLS {{?RFC8446}}
-have been used as fallbacks, when UDP is blocked for some reason.
+packet losses, the default underlying transport protocol has historically been
+UDP {{?RFC0768}}, but a large variety of other underlying transport protocols
+have been defined for various reasons (e.g., securing media exchange, or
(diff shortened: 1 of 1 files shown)


SpencerDawkins:
@mengelbart - thanks for pushing back on this one. I rephrased this text during my lightning-round review, and I think I said too much, to make my point, which was intended to be 

- we've used UDP since forever, but
- we've been adding underlying transports for nearly that long. 

So, I'm thinking that instead of 

> The Real-time Transport Protocol (RTP) [[RFC3550](https://mengelbart.github.io/rtp-over-quic-draft/draft-ietf-avtcore-rtp-over-quic.html#RFC3550)] is generally used to carry real-time media for conversational media sessions, such as video conferences, across the Internet. Since RTP requires real-time delivery and is tolerant to packet losses, the default underlying transport protocol has historically been UDP [[RFC0768](https://mengelbart.github.io/rtp-over-quic-draft/draft-ietf-avtcore-rtp-over-quic.html#RFC0768)]. More recently, DTLS [RFC9147] has been used as an underlying transport in order to secure the media exchange, and other transport protocols such as TCP [RFC9293] and TLS [RFC8446] have been used as fallbacks, when UDP is blocked for some reason.

I SHOULD have said 

> The Real-time Transport Protocol (RTP) [[RFC3550](https://mengelbart.github.io/rtp-over-quic-draft/draft-ietf-avtcore-rtp-over-quic.html#RFC3550)] is generally used to carry real-time media for conversational media sessions, such as video conferences, across the Internet. Since RTP requires real-time delivery and is tolerant to packet losses, the default underlying transport protocol has historically been UDP [[RFC0768](https://mengelbart.github.io/rtp-over-quic-draft/draft-ietf-avtcore-rtp-over-quic.html#RFC0768)], **but a large variety of other underlying transport protocols have been defined for various reasons (e.g., securing media exchange, or providing a fallback when UDP is blocked along a network path). This document describes RTP over QUIC, providing one more underlying transport protocol. The reasons for using QUIC as an underlying transport protocol are given in {{motivations}}}.**

mengelbart:
Thanks, that sounds good. I updated the PR and will merge it now.

PR #206, opened 2024-04-22
Fix 5-tuple sentence

mengelbart:
At least one *(* is missing, but since 5-tuple is explained before, I think we can just omit it.


SpencerDawkins (approved):
Oh, yes. PLEASE make this change! The repetition is painful to read.

PR #207, opened 2024-04-22
Fix target bitrate font/style

mengelbart:


SpencerDawkins (approved):
Yes. This term appears as a variable in SCReAM, but doesn't even appear in NADA, so we can't reasonably use the term in this form. Good catch!

PR #212, opened 2024-04-29
Fix a race condition cause by out of band signaling

mengelbart:
Signaling flow IDs may happen out-of-band and cause a race condition when flow IDs are received before signaling is complete.

The solution is modeled after WebTransport's handling of incoming streams and datagrams described in https://datatracker.ietf.org/doc/html/draft-ietf-webtrans-http3-09#section-4.5


SpencerDawkins (approved):
I'm slightly concerned that we're still finding "QUIC datagram(s)" in the text - I thought I'd upshifted all of those in a previous PR. Maybe I missed some?

mengelbart:
> I'm slightly concerned that we're still finding "QUIC datagram(s)" in the text - I thought I'd upshifted all of those in a previous PR. Maybe I missed some?

I think I introduced some new ones in this PR. I updated it and also applied the other suggestions you made.

PR #213, opened 2024-04-29
Apply Jörg's review comments

mengelbart:

PR #214, opened 2024-04-29
Fix streams terminology

mengelbart:
Using *STREAM frames* the same way we use *DATAGRAM frames* created some issues. For example, the stream encapsulation section could be misunderstood as *every QUIC stream frame starts with a flow identifier*, which is incorrect. I hope this PR makes it more clear.

This PR should wait for #213 before we merge it.

### DIFF
--- a/draft-ietf-avtcore-rtp-over-quic.md
+++ b/draft-ietf-avtcore-rtp-over-quic.md
@@ -1159,11 +1159,11 @@ This document describes RoQ in sufficient detail that an implementer can build a
 
 Possible directions would include
 
-* More guidance on transport for RTCP (for example, when to use QUIC streams vs. DATAGRAMs).
+* More Guidance on transport for RTCP (for example, when to use QUIC streams vs. DATAGRAMs).
 
-* More guidance on the use of real-time-friendly congestion control algorithms (for example, Copa {{Copa}}, L4S {{?RFC9330}}, etc.).
+* More Guidance on the use of real-time-friendly congestion control algorithms (for example, Copa {{Copa}}, L4S {{?RFC9330}}, etc.).
 
-* More guidance for congestion control and rate adaptation for multiple RoQ flows (whether streams or datagrams).
+* More Guidance for congestion control and rate adaptation for multiple RoQ flows (whether streams or datagrams).
 
 * Possible guidance for connection sharing between real-time and non-real-time flows, including considerations for congestion control and rate adaptation, scheduling, prioritization, and which ALPNs to use.
 
--- a/draft-ietf-avtcore-rtp-over-quic.md
+++ b/draft-ietf-avtcore-rtp-over-quic.md
@@ -243,7 +243,7 @@ Sender:
 : An endpoint that sends media in RTP packets and might send or receive RTCP packets.
 
 Stream:
-: The term "stream" is ambiguous. Without a qualifier, "stream" could refer to a QUIC STREAM frame, as defined in {{!RFC9000}}, or a series of QUIC STREAM frames in a single stream, or a series of RTP packets encapsulated in QUIC STREAM frames. If not explicitly qualified, the term "STREAM" in this document refers to a QUIC STREAM frame, and "stream" in this document refers to one or more RTP packets encapsulated in QUIC STREAM frames. This document also uses the term "RoQ stream" as a short form of "one or more RTP packets encapsulated in QUIC STREAM frames".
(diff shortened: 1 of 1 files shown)


draft-ietf-avtcore-rtp-over-quic.md
```diff
@@ -658,7 +657,7 @@ credits it will have to provide to the sending peer, and how rapidly it must pro
 Senders can also transmit RTP packets in QUIC DATAGRAMs, using
 a QUIC extension described in {{!RFC9221}}.
 DATAGRAMs can only be used if the use of the DATAGRAM extension was successfully negotiated during the QUIC handshake.
-If the QUIC extension was negotiated using a signaling protocol, but that extension was not negotiated during the resulting QUIC handshake, an endpoint can close the connection with the ROQ\_EXPECTATION\_UNMET error code.
+If the QUIC extension was negotiated using a signaling protocol, but that extension was not negotiated during the resulting QUIC handshake, an endpoint MAY close the connection with the ROQ\_EXPECTATION\_UNMET error code.
```
SpencerDawkins, line 660: Why is this changed to a normative "MAY"? But if we are thinking about normative behavior, why would an endpoint not close the connection immediately? It seems odd to notice a mismatch at signaling time, but not give the peer endpoint any indication that all is not right, until the peer endpoint tries to use the QUIC extension. My understanding of [Section 12.4 in RFC 9000](https://datatracker.ietf.org/doc/html/rfc9000#name-frames-and-frame-types) is that the result would be a connection error of type FRAME_ENCODING_ERROR, and that doesn't point directly to a signaling/negotiation mismatch, which is the real problem.

draft-ietf-avtcore-rtp-over-quic.md
```diff
@@ -899,21 +896,21 @@ QUIC connection layer to reduce the RTCP overhead.
...
-regarded as acknowledged when the QUIC Datagram frame that carried the RTP
+acknowledgments of QUIC DATAGRAM frames. In the following, an RTP packet is
+regarded as acknowledged when the QUIC DATAGRAM frame that carried the RTP
 packet was acknowledged.
 
 ## RoQ Streams {#roc-s}
 
 For RTP packets that are sent over QUIC streams, an RTP packet is considered
-acknowledged after all frames that carried fragments of the RTP packet were
+acknowledged after all frames that carried parts of the RTP packet were
```
SpencerDawkins, line 906:
```suggestion
acknowledged after all STREAM frames that carried parts of the RTP packet were
```

SpencerDawkins (approved):
If I understand correctly, we are moving from the term "congestion avoidance" to "congestion control" in this document. That's fine, but I'd suggest adding that to the Note to the Reader about this, replacing 

> Note to the Reader: the meaning of the terms "congestion control" and "rate adaptation" in the IETF community have evolved over the decades since "slow start" and "congestion avoidance" were added as mandatory to implement in TCP, in Section 4.2.2.15 of [RFC1122].  

with 

> Note to the Reader: the meaning of the terms **"congestion avoidance",** "congestion control", and "rate adaptation" in the IETF community have evolved over the decades since "slow start" and "congestion avoidance" were added as mandatory to implement in TCP, in Section 4.2.2.15 of [RFC1122].

draft-ietf-avtcore-rtp-over-quic.md
mengelbart: At the end of [section 3.2](https://mengelbart.github.io/rtp-over-quic-draft/draft-ietf-avtcore-rtp-over-quic.html#section-3.2), we say that an application can use streams instead. FRAME_ENCODING_ERROR would only happen if the QUIC stack is sending datagrams even though the datagram extension was not negotiated at connection establishment. In that case, I would expect the QUIC stack to return an error when the application is trying to send datagrams.

SpencerDawkins: > At the end of [section 3.2](https://mengelbart.github.io/rtp-over-quic-draft/draft-ietf-avtcore-rtp-over-quic.html#section-3.2), we say that an application can use streams instead. FRAME_ENCODING_ERROR would only happen if the QUIC stack is sending datagrams even though the datagram extension was not negotiated at connection establishment. In that case, I would expect the QUIC stack to return an error when the application is trying to send datagrams.

Oh, right. This is a variation of the same conversation we were having at the Hackathon in San Francisco - sorry for MY confusion. 

One thing that doesn't help me stay unconfused, is that we are talking about a specific extension here, but the text sounds more general. At the end of [Section 3.2](https://mengelbart.github.io/rtp-over-quic-draft/draft-ietf-avtcore-rtp-over-quic.html#name-rtp-with-quic-streams-quic-), we explain this more clearly than we're saying in this text. The point in Section 3.2 is "you have a decision to make. How badly do you need to use DATAGRAMs?" 

```suggestion
If the DATAGRAM extension was negotiated using a signaling protocol, but was not also negotiated during the resulting QUIC handshake, an endpoint can close the connection with the ROQ\_EXPECTATION\_UNMET error code.
```

PR #221, opened 2024-06-05
Fill out BBC gst-roq implementation section

samhurst:
As discussed on the AVTcore interim call yesterday, I've tried to fill out the gst-roq part in the Implementation Experience section.


mengelbart:
Thank you @samhurst!

PR #227, opened 2024-07-25
Add RTCP overhead considerations for streams/datagrams

mengelbart:


gloinul:
So I think there are a couple of things that needs to be considered here for the RTCP sending.

When it comes to RTCP bandwidth calculations and handling of avg_rtcp_size I think we can start considering [RFC 3890](https://datatracker.ietf.org/doc/rfc3890/) and its statement that RTCP indifference between IPv4 and IPv6 does not matter, a difference in many cases that are between 28 and 48 bytes of overhead per RTCP compound packet. I think removing the overhead completely from avg_rtcp_size will create issues in interworking with session legs that use RTP over IP/UDP etc. I think adding the actual overhead per transmission of RTCP compound packets to avg_rtcp_size will ensure that RoQ endpoints sends within more reasonable bounds when interworking. I think IP+UDP+QUIC packet headers + DATAGRAM header would be a reasonable overhead to add. If not added and RoQ endpoint have 0 overhead their RTCP transmission rate could easily become 50% or more higher. 

Worth noting is that [RFC 5506] (https://datatracker.ietf.org/doc/rfc5506/) for reduced size RTCP, i.e. without SR/RR and SDES items are calculated into avg_rtcp_size. Thus, if one sends significant amount of reduced sized RTCP packets the avg_rtcp_size value can become fairly small. Where regular RTCP packets will have RTCP compound packet sizes of larger than 100 bytes, the reduced size packets can easily be just a few 10 of bytes. 

If one doesn't have an interop case with non RoQ endpoints this would not be a problem from fairness and possible timeout perspective. The issue is that an RoQ endpoint will not know that this is the case, and explicit signalling would be needed. And that signalling may change mid session when new endpoint joins a multi-party session. So I think it would be better to include some overhead that is the average/common per transmission. 

I will note that the timeout for others parties are mostly a red-herring assuming AVPF with reasonable parameters. If the T-rr-intervall is configured to a non-zero value and which is larger than actual Td with more than a factor two (or rather the unfairness ratio between calculating entities) or so then timeout will not be an issue. For non AVPF sessions the unfairness realtionship in calculated Td must close to a factor of 5 to potentially trigger a erronous timeout of an SSRC. 

Next, I think we need to have a discussion about how one best configure RTCP sending for RoQ. 
Assuming AVPF controls I think one can actually run the RTCP transmission with t-rr-interval to how often one want regular reports and maintaing sessions memebership info. Then configure the RTCP bandwidth quite high for the session. The issue remain with dynamic changes of session bandwidth and that RTCP doesn't scale. But as RTCP will be congestion controlled in RoQ case, one at least could add a recommendation to delay scheduling of the next if there is a queue RTCP packets, possible also if there are RTP packets queue, but this I think needs a bit more thought. Depending on ones API to the QUIC stack there would actually be good to rather have an event driven transmission of RTCP so that RTP stack notice need to send RTCP packet, gives stack an size estimate, gets a notification when it can transmit and then generates the packet. This only matters if the delay between requesting transmission and actual transmission is long enough that new RTCP messages or values would make sense. However, I think limiting the RTCP if the QUIC congestion control says no would be good. The problem is that if the actual CC limit is preventing sending RTCP at least each T-rr-interval one gets into timeout country. However, I think that is fairly uncommon if T-rr-interval is 1 or more seconds, and actual RTCP bandwidth gives you a regular Td on ms scale.

draft-ietf-avtcore-rtp-over-quic.md
```diff
@@ -931,6 +933,73 @@ As noted in {{multiplexing}}, all RoQ streams and RoQ datagrams begin with a flo
...
+
+When sending RTCP packets in QUIC datagrams this implies that a packet may not
+be immediately transmitted as it is subject to queuing and multiplexing with RTP
+packets and subject to congestion control. This means that a sending timestamp
+added to an RTCP packet, e.g., in an SR packet, may differ in unforeseeable ways
+from the actual time when the packet gets sent into the network while these are
+usually fairly close to each other for RTP-over-UDP. Effectively, we have a
+application sending timestamp *t_a* and the network transmission timestamp
+*t_n*. Applications just have to be aware that RTCP does not measure the network
+level RTT but rather the application layer RTT.
```
SpencerDawkins, line 955: I'm not sure what we're telling the application (developer) here. If I thought I had *t_n*, but actually had *t_a*, my RTT estimates are going to be more conservative, aren't they? If I managed to use *t_n* for RTT estimation, doesn't that mean I'm more likely to overshoot what the receiver can accept?

draft-ietf-avtcore-rtp-over-quic.md
```diff
@@ -931,6 +933,73 @@ As noted in {{multiplexing}}, all RoQ streams and RoQ datagrams begin with a flo
 
 Because QUIC provides a capability to migrate connections for various reasons, including recovering from a path failure ({{Section 9 of !RFC9000}}), when a QUIC connection migrates, a RoQ sender has the opportunity to revisit decisions about which RTP packets are encapsulated in QUIC streams, and which RTP packets are encapsulated in QUIC DATAGRAMs. Again, RoQ receivers need to be prepared for this eventuality.
 
+## RTCP Considerations
+
+RTCP was originally defined to be used with UDP, which implies (1) that
+transmission timing is largely under the control of the application (limited
+buffering inside the stack) and (2) that the overhead, *avg_rtcp_size*, used to
```
SpencerDawkins, line 940:
```suggestion
RTCP was originally defined to be used with UDP, which implies (1) 
the only buffering present would be at the interface level, so that transmission timing is largely under the control of the application, and (2) that the overhead, *avg_rtcp_size*, used to
```

draft-ietf-avtcore-rtp-over-quic.md
```diff
@@ -931,6 +933,73 @@ As noted in {{multiplexing}}, all RoQ streams and RoQ datagrams begin with a flo
...
+compute the RTCP transmission interval could be deterministically computed by
+adding the IP asnd UDP headers. Both change when carrying RTCP over QUIC and
+they change in different ways when using QUIC streams vs. QUIC datagrams.
+
+### RTCP over QUIC datagrams {#rtcp-over-datagrams}
+
+When sending RTCP packets in QUIC datagrams this implies that a packet may not
+be immediately transmitted as it is subject to queuing and multiplexing with RTP
+packets and subject to congestion control. This means that a sending timestamp
```
SpencerDawkins, line 949:
```suggestion
packets and subject to QUIC congestion control. This means that a sending timestamp
```

draft-ietf-avtcore-rtp-over-quic.md
SpencerDawkins, line 955: I'm not sure how this is different from RT(C)P over UDP. Isn't this last sentence also true for that?

draft-ietf-avtcore-rtp-over-quic.md
```diff
@@ -931,6 +933,73 @@ As noted in {{multiplexing}}, all RoQ streams and RoQ datagrams begin with a flo
...
+because, in addition to adding the IP and UDP headers, the QUIC (short) header
+and the QUIC datagram frame header are to be considered but their sizes vary and
+it is unknown which other frames may be sent along in the same UDP packet. Any
+lower bound that can be determined could be affected by the version of QUIC
+being used.
+
+It is thus suggested that applications do _not_ attempt to include lower layer
+headers but only work with the net data rate of the codec including the RTP
+headers and of the RTCP packets, both excluding QUIC, UDP, and IP header
+overhead.
```
SpencerDawkins, line 967:
```suggestion
It is thus suggested that application developers recognize that per-RTCP packet overhead will always be an estimate, and include IP, UDP, QUIC, and DATAGRAM header sizes as a conservative heuristic.
```
Stolen from @gloinul - thank you, Magnus!

draft-ietf-avtcore-rtp-over-quic.md
```diff
@@ -931,6 +933,73 @@ As noted in {{multiplexing}}, all RoQ streams and RoQ datagrams begin with a flo
...
+For RTT computations, again, this means that the application must consider that
+it observes the application layer RTT including retransmissions, where
+retransmissions also contribute to the observed jitter.
+
+For overhead computation, retransmissions are not explicitly considered nor is
+the multiplexing with other streams.
+
+To keep the complexity under control, it is again recommended to _not_ attempt
+to include the unknown lower layer overhead and only use the data rates and
+packet sizes visible at the application (=RTP/RTCP) layer.
```
SpencerDawkins, line 987:
```suggestion
To keep the complexity under control, it is again suggested that application developers recognize that per-RTCP packet overhead will always be an estimate, and these estimates should include plausible values for IP, UDP, QUIC, and QUIC STREAM frame header sizes. 
```
I'm adapting a @gloinul comment about datagrams - please let me know if I got this wrong.

draft-ietf-avtcore-rtp-over-quic.md
```diff
@@ -931,6 +933,73 @@ As noted in {{multiplexing}}, all RoQ streams and RoQ datagrams begin with a flo
 
 Because QUIC provides a capability to migrate connections for various reasons, including recovering from a path failure ({{Section 9 of !RFC9000}}), when a QUIC connection migrates, a RoQ sender has the opportunity to revisit decisions about which RTP packets are encapsulated in QUIC streams, and which RTP packets are encapsulated in QUIC DATAGRAMs. Again, RoQ receivers need to be prepared for this eventuality.
 
+## RTCP Considerations
```
SpencerDawkins, line 936:
```suggestion
## RTCP Considerations {#RTCP-considerations}
```

draft-ietf-avtcore-rtp-over-quic.md
```diff
@@ -931,6 +933,73 @@ As noted in {{multiplexing}}, all RoQ streams and RoQ datagrams begin with a flo
...
+### Mixed operations
+
+Applications may, in principle, mix sending RTP and RTCP via streams and as
+datagrams. Doing so has unforeseeable implications on timing and reordering and
+overhead.
+
+Therefore, applications SHOULD use the same QUIC primitives for both RTP and
+RTCP. If an application uses both streams and datagrams to selectively obtain
+reliable transmission for some RTP packets but not for others, it SHOULD also
+knowingly choose which RTT observations it is interested in.
```
SpencerDawkins, line 1001:
```suggestion
Using the same QUIC primitives for both RTP and
RTCP will be safer than mixing QUIC primitives - for example, using QUIC streams to carry RTP media payloads and QUIC DATAGRAMs to carry RTCP, or vice versa. If an application uses both streams and datagrams to selectively obtain
reliable transmission for some RTP media payloads but not for others, it is strongly suggested that the application developer
knowingly choose which RTT observations they are interested in, while remaining aware of the advice included in {{RTCP-considerations}}
```

draft-ietf-avtcore-rtp-over-quic.md
```diff
@@ -931,6 +933,73 @@ As noted in {{multiplexing}}, all RoQ streams and RoQ datagrams begin with a flo
...
+
+It is thus suggested that applications do _not_ attempt to include lower layer
+headers but only work with the net data rate of the codec including the RTP
+headers and of the RTCP packets, both excluding QUIC, UDP, and IP header
+overhead.
+
+> **Editor’s Note:** For compatibility with the computation of RTP-over-UDP
+> implementations, one could consider adding just the UDP and IP headers.
+
+### RTCP over QUIC streams
```
SpencerDawkins, line 972:
```suggestion
### RTCP over QUIC streams {#rtcp-over-streams}
```

SpencerDawkins (approved):

draft-ietf-avtcore-rtp-over-quic.md
```diff
@@ -931,6 +933,73 @@ As noted in {{multiplexing}}, all RoQ streams and RoQ datagrams begin with a flo
...
+being used.
+
+It is thus suggested that applications do _not_ attempt to include lower layer
+headers but only work with the net data rate of the codec including the RTP
+headers and of the RTCP packets, both excluding QUIC, UDP, and IP header
+overhead.
+
+> **Editor’s Note:** For compatibility with the computation of RTP-over-UDP
+> implementations, one could consider adding just the UDP and IP headers.
+
```
SpencerDawkins, line 971:
```suggestion

```
(This is just removing the Editor's Note)

draft-ietf-avtcore-rtp-over-quic.md
```diff
@@ -931,6 +933,73 @@ As noted in {{multiplexing}}, all RoQ streams and RoQ datagrams begin with a flo
...
+For overhead computation, retransmissions are not explicitly considered nor is
+the multiplexing with other streams.
+
+To keep the complexity under control, it is again recommended to _not_ attempt
+to include the unknown lower layer overhead and only use the data rates and
+packet sizes visible at the application (=RTP/RTCP) layer.
+
+> **Editor’s note:** previous note also applies here but, due to
+> retransmissions, overhead is even less predictable.
```
SpencerDawkins, line 990:
```suggestion

```
This is just removing the Editor's note, which refers to a previous Editor's note that no longer exists.

SpencerDawkins (approved):
We definitely should ping @gloinul to see if I understood his comments!

draft-ietf-avtcore-rtp-over-quic.md
gloinul: Well, in most cases when running RTP/RTCP directly over UDP the only extra latency is the time it takes from having entered the application layer timestamp until an IP/UDP packet with the payload can be created. That is normally quite short when one is running directly on UDP sockets.

draft-ietf-avtcore-rtp-over-quic.md
mengelbart: I think this should also include QUIC STREAM frames.

mengelbart: oh, no, sorry, STREAMs are covered below...

PR #232, opened 2024-11-12
Added a paragraph about RTCP over RoQ

nils-ohlmeier:
After reading the draft and starting with an implementation I was left wondering how I should transport RTCP over QUIC. This PR adds a paragraph which attempts to clarify the available options.

### DIFF
--- a/draft-ietf-avtcore-rtp-over-quic.md
+++ b/draft-ietf-avtcore-rtp-over-quic.md
@@ -742,6 +742,14 @@ RTP retransmissions can be done in the same RTP session or in a
 different RTP session {{!RFC4588}} and the flow identifier MUST be set to the
 flow identifier of the RTP session in which the retransmission happens.
 
+## RTCP
+
+The same encapsulation as described above for RTP packets can also be used to carry RTCP packets back from the receiver to the sender. Similar to the RTP packet RTCP can use datagrams or streams as well.
+If a receiver sends aggregated RTCP reports for multiple RTP streams the flow identifier no longer machtes the flow identifier for a single RTP stream. Thus the sender always needs to inspect the received RTCP packet independent of the flow identifier used to the RTCP flow to determine to which of the RTP flows the received packets apply.
+This is also the reason why bidirectional streams are not recommendend, as the the received RTCP packets would not necessarily apply to the same RTP stream being send on the same flow.
+Further more using a bidirectional stream could result in a situation where the sender has closed its side of the QUIC stream, but the receiver continues to send RTCP in the opposite direction.
+Thus is makes more sense if sender and receiver agree on one or multiple unidictional flows from the receiver to the sender to transport any RTCP messages back to the sender.
+
 # Connection Shutdown
 
 Either endpoint can close the connection for any of a variety of reasons. If one of the
--- a/draft-ietf-avtcore-rtp-over-quic.md
+++ b/draft-ietf-avtcore-rtp-over-quic.md
@@ -744,7 +744,7 @@ flow identifier of the RTP session in which the retransmission happens.
 
 ## Encapsulation Considerations for RTCP
 
-The same encapsulation as described above for RTP packets can also be used to carry RTCP packets back from the receiver to the sender. Both RTP and RTCP can be transported in either QUIC DATAGRAM frames or QUIC STREAM frames. 
+The same encapsulation as described above for RTP packets can also be used to carry RTCP packets back from the receiver to the sender. Both RTP and RTCP can be transported in either QUIC DATAGRAM frames or QUIC STREAM frames.
 If a receiver sends aggregated RTCP reports for multiple RTP streams, the flow identifier no longer matches the flow identifier for a single RTP stream. Thus the sender always needs to inspect the received RTCP packet independent of the flow identifier used to the RTCP flow to determine to which of the RTP flows the received packets apply.
 This is also the reason why bidirectional streams are not allowed, as the received RTCP packets would not necessarily apply to the same RTP stream being sent on the same flow.
 In addition, allowing a bidirectional stream could result in a situation where the sender has closed its side of the QUIC stream, but the receiver continues to send RTCP in the opposite direction.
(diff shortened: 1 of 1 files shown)


draft-ietf-avtcore-rtp-over-quic.md
```diff
@@ -742,6 +742,14 @@ RTP retransmissions can be done in the same RTP session or in a
...
 flow identifier of the RTP session in which the retransmission happens.
 
+## RTCP
+
+The same encapsulation as described above for RTP packets can also be used to carry RTCP packets back from the receiver to the sender. Similar to the RTP packet RTCP can use datagrams or streams as well.
+If a receiver sends aggregated RTCP reports for multiple RTP streams the flow identifier no longer machtes the flow identifier for a single RTP stream. Thus the sender always needs to inspect the received RTCP packet independent of the flow identifier used to the RTCP flow to determine to which of the RTP flows the received packets apply.
+This is also the reason why bidirectional streams are not recommendend, as the the received RTCP packets would not necessarily apply to the same RTP stream being send on the same flow.
+Further more using a bidirectional stream could result in a situation where the sender has closed its side of the QUIC stream, but the receiver continues to send RTCP in the opposite direction.
+Thus is makes more sense if sender and receiver agree on one or multiple unidictional flows from the receiver to the sender to transport any RTCP messages back to the sender.
```
mengelbart, line 751:
```suggestion
Thus it makes more sense if the sender and receiver agree on one or multiple unidirectional streams to transport any RTCP messages.
```

mengelbart: I think it makes sense to put this more generally since RTCP may go both ways.

mengelbart (commented):
I added one comment/suggestion, but other than that, it looks good to me. Thanks!

nils-ohlmeier:
Thanks for the suggestion. Yes that wording is better.

draft-ietf-avtcore-rtp-over-quic.md
```diff
@@ -742,6 +742,14 @@ RTP retransmissions can be done in the same RTP session or in a
 different RTP session {{!RFC4588}} and the flow identifier MUST be set to the
 flow identifier of the RTP session in which the retransmission happens.
 
+## RTCP
+
+The same encapsulation as described above for RTP packets can also be used to carry RTCP packets back from the receiver to the sender. Similar to the RTP packet RTCP can use datagrams or streams as well.
```
SpencerDawkins, line 747:
```suggestion
The same encapsulation as described above for RTP packets can also be used to carry RTCP packets back from the receiver to the sender. Both RTP and RTCP can be transported in either QUIC DATAGRAM frames or QUIC STREAM frames. 
```

draft-ietf-avtcore-rtp-over-quic.md
```diff
@@ -742,6 +742,14 @@ RTP retransmissions can be done in the same RTP session or in a
...
+If a receiver sends aggregated RTCP reports for multiple RTP streams the flow identifier no longer machtes the flow identifier for a single RTP stream. Thus the sender always needs to inspect the received RTCP packet independent of the flow identifier used to the RTCP flow to determine to which of the RTP flows the received packets apply.
+This is also the reason why bidirectional streams are not recommendend, as the the received RTCP packets would not necessarily apply to the same RTP stream being send on the same flow.
+Further more using a bidirectional stream could result in a situation where the sender has closed its side of the QUIC stream, but the receiver continues to send RTCP in the opposite direction.
+Thus it makes more sense if the sender and receiver agree on one or multiple unidirectional streams to transport any RTCP messages.
+
```
SpencerDawkins, line 752: I'm not sure whether we need the text about bidirectional streams, using them would be a protocol error.  That's from [Section 5.2](https://mengelbart.github.io/rtp-over-quic-draft/draft-ietf-avtcore-rtp-over-quic.html#name-quic-streams) - bidirectional streams are a ROQ_STREAM_CREATION_ERROR.

I think we still need the final sentence in the new text. @mengelbart, what do you think?

```suggestion
If a receiver sends aggregated RTCP reports for multiple RTP streams, the flow identifier no longer matches the flow identifier for a single RTP stream. Thus the sender always needs to inspect the received RTCP packet independent of the flow identifier used to the RTCP flow to determine to which of the RTP flows the received packets apply.
This is also the reason why bidirectional streams are not allowed, as the received RTCP packets would not necessarily apply to the same RTP stream being sent on the same flow.
In addition, allowing a bidirectional stream could result in a situation where the sender has closed its side of the QUIC stream, but the receiver continues to send RTCP in the opposite direction.
Thus it makes more sense if the sender and receiver agree on one or multiple unidirectional streams to transport any RTCP messages.

```

draft-ietf-avtcore-rtp-over-quic.md
```diff
@@ -742,6 +742,14 @@ RTP retransmissions can be done in the same RTP session or in a
 different RTP session {{!RFC4588}} and the flow identifier MUST be set to the
 flow identifier of the RTP session in which the retransmission happens.
 
+## RTCP
```
SpencerDawkins, line 745:
```suggestion
## Encapsulation Considerations for RTCP
```

SpencerDawkins (approved):
I had a couple of editorial changes, but my most significant comment is about the new text on bidirectional streams. I THINK that's a RoQ protocol error (details in my comments starting on line 748).

draft-ietf-avtcore-rtp-over-quic.md
mengelbart: I agree that it is a protocol violation to use bidirectional streams, but I think it is helpful to have this explanation as to why it doesn't make sense to use bidirectional streams to relate RTCP packets to RTP packets. Maybe change the words `not recommendend` to something else because it is not only not recommend, but forbidden to use bidirectional streams?

nils-ohlmeier: @mengelbart Spencer changed the "not recommended" already into "not allowed". Are we happy with that, or should I try to come up with a better phrasing?

mengelbart: Sorry for the slow response. I'm happy with applying Spencer's suggestion and then merging this.

SpencerDawkins (approved):
I think this is ready to add @nils-ohlmeier to the Acks section. Thank you for sending text! I hope everyone does ...

draft-ietf-avtcore-rtp-over-quic.md
```diff
@@ -742,6 +742,14 @@ RTP retransmissions can be done in the same RTP session or in a
 different RTP session {{!RFC4588}} and the flow identifier MUST be set to the
 flow identifier of the RTP session in which the retransmission happens.
 
+## Encapsulation Considerations for RTCP
+
+The same encapsulation as described above for RTP packets can also be used to carry RTCP packets back from the receiver to the sender. Both RTP and RTCP can be transported in either QUIC DATAGRAM frames or QUIC STREAM frames.
```
mengelbart, line 747:
```suggestion
The same encapsulation as described above for RTP packets can also be used to carry RTCP packets back from the receiver to the sender. Both RTP and RTCP can be transported in either QUIC DATAGRAM frames or QUIC STREAM frames.
```
This may seem nitpicky, but I think it would make the build fail (locally, it does). I can't run actions on the PR, though, because it still has an old version of the Github action, which Github refuses to run... Let's hope the build works correctly when I merge it to the main branch.

PR #234, opened 2025-02-06
Update acknowledgment section

mengelbart:
Should we add affiliations to the implementer's acknowledgments?


SpencerDawkins (approved):
Hi, @mengelbart, this looks fine. Please proceed! 

And I thought I had responded to say that we probably don't need to add affiliations in the Acks section. We ask for that on blue sheets, but that's only so we can disambiguate five Fred Smiths who are active at any one time, for purposes of NomCom eligibility and (more rarely) investigations on submarine patents. 

If we were ever going to start doing that in RFCs, we probably shouldn't do that during a tech industry economic downturn! :unamused:

SpencerDawkins:
Actually, I can just merge this as the other reviewer. Silly me!

